PR #12, opened 2017-10-09
Add a sample

rynowak:
FYI only - will merge when the build passes

Adding a (basic) sample and a (bad) implementation for us to iterate upon


rynowak:
There's no travis and appveyor #YOLO

PR #16, opened 2017-10-11
Add named clients and options and caching

rynowak:
Adds the ability to create named clients
- built on top of named options
- this means introducing an options type

Also adds caching of handlers

----------------

The options type looks a little wonky but humour me for a second here.

Fundamentally I think it's the right strategy for us to cache *handlers*
not clients. The handler is the thing that actually does connection
management, and conveniently they are opaque and immutable once
constructed.

The HttpClient on the other hand *is* mutable. So if you do something
like dispose it, or modify the big public mutable things, you do so for
the rest of the callers in the system.

I think that caching handlers, but not clients gets us everything we
want. In addition using the HttpClient with either pattern (keep forever
or dispose) will 'just work' without any complicated code or drawbacks
for users.

------

So after establishing that, we need to a way to express both deferred
operations that build a 'handler' pipeline, and deferred operations that
modify an HttpClient - and that's what it looks like on the options
type.

We should add extension methods or maybe another builder if necessary to
smooth it over, but I think we'll get pretty far with what I have until
we start tying in 3rd party libraries.


samples/HttpClientFactorySample/Program.cs
```diff
@@ -19,19 +19,23 @@ public class Program
...
         {
             var services = new ServiceCollection()
-                .AddHttpClient()
+                .AddHttpClient("github", c =>
+                {
+                    c.BaseAddress = new Uri("https://api.github.com/");
+
+                    c.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json"); // Github API versioning
+                    c.DefaultRequestHeaders.Add("User-Agent", "HttpClientFactory-Sample"); // Github requires a user-agent
+                })
```
rynowak: I think that most of the 'common' stuff we want is right on the `HttpClient` - until we start talking about 3rd party libraries

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
```diff
@@ -1,17 +1,77 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using System.Collections.Concurrent;
+using System.Threading;
 
 namespace Microsoft.Extensions.Http
 {
-    // This is just a starter implementation. It's obviously wrong on purpose.
-    // The plan is to get a basic sample in first, and improve this along with the sample.
     internal class DefaultHttpClientFactory : HttpClientFactory
```
rynowak: This class:
- Creates the actual client 
- Does caching
- Applies the options boilerplate

The only real reason to replace/extend this is if you want a different policy for how to create the client and do caching.

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
```diff
@@ -1,17 +1,77 @@
...
+            {
+                options.HttpClientActions[i](client);
+            }
+
+            return client;
+        }
+
+        private HttpMessageHandler CreateHandler(string name)
+        {
+            var builder = _services.GetRequiredService<HttpMessageHandlerBuilder>();
```
rynowak: Using DI as a factory.

src/Microsoft.Extensions.Http/DefaultHttpMessageHandlerBuilder.cs
```diff
@@ -0,0 +1,53 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+
+namespace Microsoft.Extensions.Http
+{
+    internal class DefaultHttpMessageHandlerBuilder : HttpMessageHandlerBuilder
```
rynowak: This needs some tests, waiting for feedback on design first

src/Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs
```diff
@@ -15,8 +16,34 @@ public static IServiceCollection AddHttpClient(this IServiceCollection services)
...
             }
 
+            services.AddOptions();
+
+            services.AddTransient<HttpMessageHandlerBuilder, DefaultHttpMessageHandlerBuilder>();
             services.AddSingleton<HttpClientFactory, DefaultHttpClientFactory>();
             return services;
         }
+
+        public static IServiceCollection AddHttpClient(this IServiceCollection services, string name, Action<HttpClient> configureClient)
```
rynowak: Not exposing an o/l with `Action<HttpClientFactoryOptions>` because the actual options type is pretty gross and not really intended for user code.

src/Microsoft.Extensions.Http/HttpClientBuilder.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    /// <summary>
+    /// A builder abstraction for configuring <see cref="HttpMessageHandler"/> instances.
+    /// </summary>
+    /// <remarks>
+    /// The <see cref="HttpMessageHandlerBuilder"/> is registered in the service collection as
+    /// a transient service. Callers should retrieve a new instance for each <see cref="HttpMessageHandler"/> to
+    /// be created. Implementors should expect each instance to be used a single time.
+    /// </remarks>
+    public abstract class HttpMessageHandlerBuilder
```
rynowak: This is a replaceable abstraction. You can act as a 'coordinator' for the handlers, or configure your own set of default handlers.

src/Microsoft.Extensions.Http/HttpClientBuilder.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// Gets a list of additional <see cref="DelegatingHandler"/> instances used to configure an
+        /// <see cref="HttpClient"/> pipeline.
+        /// </summary>
+        public abstract IList<DelegatingHandler> AdditionalHandlers { get; }
+
+        /// <summary>
+        /// Creates an <see cref="HttpMessageHandler"/>.
+        /// </summary>
+        /// <returns></returns>
+        public abstract HttpMessageHandler Build();
```
rynowak: Most of the code that's in our implementation could be moved here as a protected static. It's boilerplate and error checking.

src/Microsoft.Extensions.Http/HttpClientBuilder.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <summary>
+        /// Gets a list of additional <see cref="DelegatingHandler"/> instances used to configure an
+        /// <see cref="HttpClient"/> pipeline.
+        /// </summary>
+        public abstract IList<DelegatingHandler> AdditionalHandlers { get; }
+
+        /// <summary>
+        /// Creates an <see cref="HttpMessageHandler"/>.
+        /// </summary>
+        /// <returns></returns>
```
rynowak: blep

src/Microsoft.Extensions.Http/HttpClientFactoryOptions.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        /// <summary>
+        /// Gets a list of operations used to configure an <see cref="HttpMessageHandlerBuilder"/>.
+        /// </summary>
+        public IList<Action<HttpMessageHandlerBuilder>> HandlerBuilderActions { get; } = new List<Action<HttpMessageHandlerBuilder>>();
+
+        /// <summary>
+        /// Gets a list of operations used to configure an <see cref="HttpClient"/>.
+        /// </summary>
+        public IList<Action<HttpClient>> HttpClientActions { get; } = new List<Action<HttpClient>>();
```
rynowak: This type is actually kinda gross - but I think this is the best approach. See the commit comments

src/Microsoft.Extensions.Http/HttpClientBuilder.cs
lecaillon: File and class names are different.

davidfowl: Hmm, there's something similar in HttpAbstractions with the `IAppliactonBuilder` and the `IApplicaitonBuilderFactory`.

test/Microsoft.Extensions.Http.Test/DefaultHttpClientFactoryTest.cs
```diff
@@ -1,23 +1,135 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.Extensions.DependencyInjection;
```
davidfowl: Fix your VS settings to put System stuff first

samples/HttpClientFactorySample/Program.cs
```diff
@@ -19,19 +19,23 @@ public class Program
         public static async Task Run()
         {
             var services = new ServiceCollection()
-                .AddHttpClient()
+                .AddHttpClient("github", c =>
+                {
+                    c.BaseAddress = new Uri("https://api.github.com/");
```
davidfowl: This is nice because it's the existing API. We don't have to worry about keeping things in sync. What does consumption look like? Do you do a using around the created http client since we're going to be caching handlers?

src/Microsoft.Extensions.Http/DefaultHttpMessageHandlerBuilder.cs
davidfowl: I'm not sure about a few things:
- Abstract base class (come at me bro)
- The having the primary handler be settable? Why isn't it a ctor parameter? Is that for testing?

rynowak: > Abstract base class (come at me bro)

Per discussion this will be an interface - we think the API we need is simple and will not grow over time

> The having the primary handler be settable? 

Yes. Some of the handlers have configuration on them as well. If you care about setting those deets you need to be able to replace the handler. I think making it not settable raises more questions then answers

>Why isn't it a ctor parameter? 

Constructors don't compose. This class comes from DI anyway, that would be really complicated and not add anything.

> Is that for testing?

We'll probably use it for testing, but it's not **only** for testing

src/Microsoft.Extensions.Http/HttpClientBuilder.cs
rynowak: Yeah I think it's a pivot here that we could use DI as the factory vs create a factory abstraction. The thing is that we don't pass any parameters to the handler builder and I don't anticipate that we will - it's already surrounded by a composition mechanism.

I think using DI for a factory is the simple/right approach. If you think this is too daring let's talk more.

samples/HttpClientFactorySample/Program.cs
rynowak: >This is nice because it's the existing API.

Yeah, I'm happy about how this worked out.

> What does consumption look like? 

Not sure what you mean. What's there is there in the sample. We've had some offline discussions about usage in MVC already but that's not part of this work-item.

> Do you do a using around the created http client since we're going to be caching handlers?

You can, you don't really have to. The only thing the using would do is cancel any requests that are going on.

src/Microsoft.Extensions.Http/HttpClientBuilder.cs
rynowak: > File and class names are different.

Thanks, will fix.

rynowak:
Updated this with docs, tests and some cleanup based on feedback. I consider this pretty much ready for a final review.

rynowak:
@anurse - want to be my PR buddy for this repo? (best friends?)

src/Microsoft.Extensions.Http/IHttpClientFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// <para>
+        /// The default <see cref="IHttpClientFactory"/> implementation may cache the underlying
+        /// <see cref="HttpMessageHandler"/> instances to improve performance.
+        /// </para>
+        /// <para>
+        /// Callers are also free to mutate the returned <see cref="HttpClient"/> instance's public properties
+        /// as desired.
+        /// </para>
+        /// </remarks>
+        HttpClient CreateClient(string name);
```
rynowak: @PureKrome eat your ❤️ out

rynowak: I realise this is not the official docs - yet - but its a start

PureKrome: shocked

excited

thank you


- Thank you, thank you, thank you :)
- Thanks for thinking of me, also  ❤️  🥇

davidfowl:
@rynowak can you add an example of a logging http client "middleware"?

rynowak:
> can you add an example of a logging http client "middleware"?

I can add one but it will be a fake. This is a `.Extensions` repo - it's a layer below all the HTTP stuff.

anurse:
> want to be my PR buddy for this repo? (best friends?) ( unity?)

davidfowl:
> I can add one but it will be a fake. This is a .Extensions repo - it's a layer below all the HTTP stuff.

I mean in a sample not in the source. I'd like to see the typical pattern for message handlers that need to get services from DI. I'd also like to see how the extension methods look (UseLogging etc).

rynowak:
>  I'd like to see the typical pattern for message handlers that need to get services from DI. I'd also like to see how the extension methods look (UseLogging etc).

Cool, that will be the next PR.

src/Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs
```diff
@@ -2,20 +2,71 @@
...
+        /// <remarks>
+        /// <para>
+        /// <see cref="HttpClient"/> instances that apply the provided configuration can be retrieved using 
+        /// <see cref="IHttpClientFactory.CreateClient(string)"/> and providing the matching name.
+        /// </para>
+        /// <para>
+        /// Use <see cref="Options.Options.DefaultName"/> as the name to configure the default client.
+        /// </para>
+        /// </remarks>
+        public static IServiceCollection AddHttpClient(this IServiceCollection services, string name, Action<HttpClient> configureClient)
```
anurse: Am I correct in seeing that we don't currently have an overload that allows you to set up the `HttpMessageHandlerBuilder`?

rynowak: Correct. That's not part of this PR

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
```diff
@@ -1,17 +1,77 @@
...
-    // The plan is to get a basic sample in first, and improve this along with the sample.
-    internal class DefaultHttpClientFactory : HttpClientFactory
+    internal class DefaultHttpClientFactory : IHttpClientFactory
     {
-        public override HttpClient CreateClient()
+        private readonly IServiceProvider _services;
+        private readonly IOptionsMonitor<HttpClientFactoryOptions> _optionsMonitor;
+
+        // Lazy, because we're using a subtle pattern here to ensure that only one instance of
+        // HttpMessageHandler is created for each name.
```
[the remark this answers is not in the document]
rynowak: Yes, and I want to avoid running `_valueFactory` since it's running user code. Predicatable guarantee is better

src/Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs
anurse: Cool, just making sure I wasn't mis-reading things.

PR #22, opened 2017-10-25
Add logging and recipe for configuring handlers

rynowak:
This change adds logging message handlers (the outermost handler and
innermost handler).

The outermost handler is responsible for creating a
scope and capturing details about the logical operation. The innermost
handler logs the details of the actual request sent over the write. This
is important when you consider things like service discovery, auth or
caching as message handlers. The request the user crafted is not what
will be sent.

This is hardcoded into the system and happens after user code interacts
with the HttpMessageHandlerBuilder so that it's harder to mess up. If
you don't like it you can unregister the policy.

-------

Additionally there are some Add* methods for adding handlers using the
HttpMessageHandlerBuilder.

-------

Also moved the validation of the builder's state into a protected static
so that it's possible to reuse when subclassing the builder. This is
boilerplate that we expect anyone to want.


samples/HttpClientFactorySample/Program.cs
```diff
@@ -23,7 +26,16 @@ public class Program
                     c.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json"); // Github API versioning
                     c.DefaultRequestHeaders.Add("User-Agent", "HttpClientFactory-Sample"); // Github requires a user-agent
                 })
-                .AddLogging()
+
+                // Retry requests to github using our retry handler
+                .AddHttpMessageHandlerPolicy("github", DefaultStages.Reliability, b => b.AdditionalHandlers.Add(new RetryHandler()))
```
rynowak: Adding a policy in the `Reliability` stage means that this lambda will run before our 'logging' policy. 

Then our logging policy will add a handler at the beginning and end to make a delicious logging sandwich 🍞

samples/HttpClientFactorySample/Program.cs
```diff
@@ -23,7 +26,16 @@ public class Program
...
+
+                // Add console logger so we can see all the logging produced by the client by default.
+                .AddLogging(b => 
+                {
+                    b.AddFilter((category, level) => true); // Spam the world with logs.
```
rynowak: I may have oversold the amount of logging that happens in this PR.

samples/HttpClientFactorySample/Program.cs
```diff
@@ -40,6 +52,37 @@ public class Program
             var data = await response.Content.ReadAsAsync<JObject>();
             Console.WriteLine("Response data:");
             Console.WriteLine(data);
+
+            Console.WriteLine("Press the ANY key to exit...");
+            Console.ReadKey();
+        }
+
+        private class RetryHandler : DelegatingHandler
```
rynowak: Just a simple thing for the sample

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
```diff
@@ -66,9 +67,20 @@ private HttpMessageHandler CreateHandler(string name)
...
+                .OrderBy(group => options.HttpMessageHandlerPolicies.Stages.IndexOf(group.Key))
+                .Where(group => options.HttpMessageHandlerPolicies.Stages.Contains(group.Key));
+
+            foreach (var stage in stages)
             {
-                options.HandlerBuilderActions[i](builder);
+                foreach (var policy in stage)
+                {
+                    policy.Apply(builder);
+                }
```
rynowak: This is how the stages work. 

The fact that we ignore policies without a matching stage is a feature not a bug. You can easily take out things you don't want to run this way. We don't expect adding a new stage to users do often.

src/Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryOptionsSetup.cs
```diff
@@ -0,0 +1,60 @@
...
+            if (options == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            options.HttpMessageHandlerPolicies.Stages.Add(DefaultStages.Default);
+            options.HttpMessageHandlerPolicies.Stages.Add(DefaultStages.Reliability);
+            options.HttpMessageHandlerPolicies.Stages.Add(DefaultStages.ServiceDiscovery);
+            options.HttpMessageHandlerPolicies.Stages.Add(DefaultStages.Security);
+            options.HttpMessageHandlerPolicies.Stages.Add(DefaultStages.Diagnostics);
```
rynowak: This order makes the most sense to me. 

I personally list this approach because I think we need to try and find a consensus ordering of these kinds of policies and recommend that by default. Consider that maybe 4-5 packages deliver policies you want, and some of them are 3rd party. 

That's what I have in mind by doing things like this.

It would be **techically** easy to just assign everything a numeric order. But it may be harder to get the E2E right without a lot of tweaking.

src/Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs
```diff
@@ -24,10 +27,21 @@ public static IServiceCollection AddHttpClient(this IServiceCollection services)
                 throw new ArgumentNullException(nameof(services));
             }
 
+            services.AddLogging();
```
rynowak: @davidfowl fite me irl

src/Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs
```diff
@@ -69,5 +83,79 @@ public static IServiceCollection AddHttpClient(this IServiceCollection services,
 
             return services;
         }
+
+        public static IServiceCollection AddHttpMessageHandlerPolicy(this IServiceCollection services, string name, IHttpMessageHandlerPolicy policy)
```
rynowak: Still waiting for `Configure<TOptions, TService>` to make this even better

src/Microsoft.Extensions.Http/HttpMessageHandlerPolicyCollection.cs
```diff
@@ -0,0 +1,39 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+namespace Microsoft.Extensions.Http
+{
+    public class HttpMessageHandlerPolicyCollection : Collection<IHttpMessageHandlerPolicy>
+    {
+        public IList<string> Stages { get; } = new List<string>();
```
rynowak: This could probably be it's own list on options. I wanted to see what would happen if I did this, now I realise that it's **super** wacky #YOLO

src/Microsoft.Extensions.Http/Logging/LoggingHttpMessageHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using System;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.Http.Logging
+{
+    public class LoggingHttpMessageHandler : DelegatingHandler
```
rynowak: Use the power of imagination to pretend that there's actually some logging here.

Why two handlers? Well the 'outer' handler makes a scope and will log information about the logical operation. The 'inner' handler (this class) will log when an HTTP request goes over the wire, which may not happen in all cases.

src/Microsoft.Extensions.Http/Logging/LoggingHttpMessageHandlerPolicy.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        public void Apply(HttpMessageHandlerBuilder builder)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            // We want all of our logging message to show up as-if they are coming from HttpClient
+            var logger = _loggerFactory.CreateLogger<HttpClient>();
```
rynowak: I'm taking a bold stance here.

src/Microsoft.Extensions.Http/Logging/LoggingScopeHttpMessageHandler.cs
```diff
@@ -0,0 +1,96 @@
...
+                _logger.LogInformation("Processing Request {HttpMethod} to {Uri}", request.Method, request.RequestUri.AbsoluteUri);
+                var response = await base.SendAsync(request, cancellationToken);
+                return response;
+            }
+
+
+                
+        }
+
+        private struct Scope : IReadOnlyList<KeyValuePair<string, object>>
```
rynowak: I started actually implementing logging, but decided that I want to get feedback on the options/startup thing first. I'll want to solicit feedback about diagnostics next.

davidfowl:
Why do we need stages? Why are we diverging from a very simple middleware pipeline?

src/Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryOptionsSetup.cs
davidfowl: I'm struggling to see why we need this IIS style pipeline for http client. What concrete problems are we solving by pre-defining those very opinionated stages into the core abstraction?

anurse: Do we expect users to have handlers brought in from random places outside their control? That seems to be the major argument in favor of pipeline stages (it's why IIS has them, for example). I'd envisioned this feature much like middleware, where the app remains in control of the order of everything via the set-up methods. We don't do this for inbound middleware so it seems weird that "outbound" middleware would do it.

davidfowl: It also assumes we know what the stages should be and this seems completely premature at this point. Stages can also be built on top of a linear pipeline (see katana stage markers)

samples/HttpClientFactorySample/Program.cs
davidfowl: This goes away right?

rynowak: > DefaultStages.Reliability 

That part does. What do you mean specifically?

rynowak:
Updated this with a simpler design approach and basic logging.

I don't want to dive deep on the logging right now as I want to open an issue and have a bigger discussion (involving people not at Microsoft 👍) to get feedback about logging before we try to do a really good jaerb.

Here's what it looks like now with scopes on:
```
info: System.Net.Http.HttpClient[100]
      => HTTP GET https://api.github.com/
      Start processing HTTP request GET https://api.github.com/

// Use your imagination and pretend there is service discovery here

info: System.Net.Http.HttpClient[200]
      => HTTP GET https://api.github.com/
      Sending HTTP request GET https://api.github.com/
info: System.Net.Http.HttpClient[201]
      => HTTP GET https://api.github.com/
      Recieved HTTP response GET https://api.github.com/ - OK

// Use your imagination and pretend there is a retry policy here for when this throws (or similar).

info: System.Net.Http.HttpClient[101]
      => HTTP GET https://api.github.com/
      End processing HTTP request GET https://api.github.com/ - OK
```

src/Microsoft.Extensions.Http/DefaultHttpMessageHandlerBuilder.cs
```diff
@@ -9,47 +9,32 @@ namespace Microsoft.Extensions.Http
...
+        private readonly IHttpMessageHandlerBuilderPostInitializer _postInitializer;
+
+        public DefaultHttpMessageHandlerBuilder()
+        {
+        }
+
+        public DefaultHttpMessageHandlerBuilder(IHttpMessageHandlerBuilderPostInitializer postInitializer)
+        {
+            _postInitializer = postInitializer;
+        }
```
rynowak: The big pivot here for the first iteration is to add a 'post-initializer' which is where we hardcode logging and diagnostics. This way you don't casually break our diagnostics, you have to try if you want to do that.

If you do want to break our diagnostics, you can unregister or replace the post-initializer

src/Microsoft.Extensions.Http/DefaultHttpMessageHandlerBuilderPostInitializer.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using System;
+using System.Net.Http;
+using Microsoft.Extensions.Http.Logging;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.Http
+{
+    // Internal so we can change the requirements without breaking changes.
+    internal class DefaultHttpMessageHandlerBuilderPostInitializer : IHttpMessageHandlerBuilderPostInitializer
```
rynowak: open to suggestions about the name of the interface. Intentionally a scary name right now because halloween and I don't expect many people to implement these

src/Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs
```diff
@@ -69,5 +82,116 @@ public static IServiceCollection AddHttpClient(this IServiceCollection services,
...
+        /// <see cref="IHttpClientFactory"/>.
+        /// </param>
+        /// <returns>The <see cref="IServiceCollection"/>.</returns>
+        /// <remarks>
+        /// <para>
+        /// The <paramref name="configureClient"/> delegate will be used to configure all <see cref="HttpClient"/> instances
+        /// produced by the <see cref="IHttpClientFactory"/>.
+        /// </para>
+        /// </remarks>
+        public static IServiceCollection AddGlobalHttpClient(this IServiceCollection services, Action<HttpClient> configureClient)
```
rynowak: I'm going with the approach right now of having a series of extensions on the `IServiceCollection` directly instead of an `IHttpClientFactoryBuilder` because I want to see if it's necessary.

 If we feel like this isn't scaling as we add more functionality we will pivot.

rynowak: Global = applies to all clients regardless of name

src/Microsoft.Extensions.Http/HttpMessageHandlerBuilder.cs
```diff
@@ -35,5 +37,52 @@ public abstract class HttpMessageHandlerBuilder
         /// <see cref="AdditionalHandlers"/>.
         /// </returns>
         public abstract HttpMessageHandler Build();
+
+        protected internal static HttpMessageHandler CreateHandlerPipeline(HttpMessageHandler primaryHandler, IEnumerable<DelegatingHandler> additionalHandlers)
```
rynowak: Made this protected static so you can subclass this without too much friction. It's boilerplate, and non-trival to get right yourself.

src/Microsoft.Extensions.Http/Logging/LoggingHttpMessageHandler.cs
```diff
@@ -0,0 +1,68 @@
...
+        protected async override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            if (request == null)
+            {
+                throw new ArgumentNullException(nameof(request));
+            }
+
+            // Not using a scope here because we always expect this to be at the end of the pipeline, thus there's
+            // not really anything to surround.
+            Log.RequestStart(_logger, request);
```
rynowak: I really prefer this to the extension methods that are common. This way you don't accidentally use logging that's intended for another component. We've legitimately had bugs like this 4-5 times in MVC.

rynowak: Does anyone mind?

samples/HttpClientFactorySample/Program.cs
davidfowl: Thats what I meant, I pulled the branch and realized it doesn't build so that just needed to be updated. I'm also not a fan of the top level method on service collection to add message handlers. I'm wondering if a builder pattern would be better.

src/Microsoft.Extensions.Http/DefaultHttpMessageHandlerBuilderPostInitializer.cs
davidfowl: This feels exactly like an IStartupFilter to me. Why wouldn't we follow that pattern? Is there anything fundamental about it that wouldn't work? Today we have a built in `IStartupFilter` that sets up the DI request scope. Its very similar to this.

src/Microsoft.Extensions.Http/Logging/LoggingHttpMessageHandler.cs
davidfowl: > This way you don't accidentally use logging that's intended for another component.

Interesting. Is that because the extension methods were public? Would making them internal just solve that?

anurse: Another thought I had when doing this somewhere else was putting the extensions on `ILogger<T>` (for a specific `T`). Doesn't necessarily help you here though.

This seems OK to me. It'd be better if it could still be an extension method but C# doesn't allow that.

anurse: > Would making them internal just solve that?

That would only keep you from crossing assembly boundaries. There are often multiple categories of `ILogger` within an assembly, which gets kinda gross.

src/Microsoft.Extensions.Http/DefaultHttpMessageHandlerBuilderPostInitializer.cs
anurse: Would we want to support more than one though? Is this a "real" extension point or just a hack we're doing to ensure we can put stuff in that wraps the pipeline?

And, if we're hard coding this in our core package, maybe we should just go the full distance and always do the logging stuff.

src/Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs
anurse: Given that most 3rd-party extensions will hang off the `HttpMessageHandlerBuilder` I think that's probably OK.

src/Microsoft.Extensions.Http/Logging/EventIds.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.Http.Logging
+{
+    internal static class EventIds
```
anurse: These are only used in the `Log` static method. Why not just have them be constants there? Also, use `nameof`?

src/Microsoft.Extensions.Http/Logging/EventIds.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        // Logging done by LoggingScopeHttpMessageHandler - this surrounds the whole pipeline
+        public static readonly EventId RequestPipelineStart = new EventId(100, "RequestPipelineStart");
```
anurse: Ugh. I dislike artificially spacing out Event Ids. Just leads to weird patterns and trying to assign meaning to things that are really just arbitrary integers identifying an event. In this case, it's not likely we'll overrun these regions but it still just seems unnecessary. The names are descriptive enough and if we document it, we'll just have a list of IDs and what they map to and I don't think users will care if they aren't "nicely ordered"

src/Microsoft.Extensions.Http/Logging/LoggingHttpMessageHandler.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            private static readonly Action<ILogger, HttpMethod, Uri, Exception> _requestStart;
+            private static readonly Action<ILogger, HttpMethod, Uri, HttpStatusCode, Exception> _requestEnd;
+
+            private static readonly string MessageRequestStart = "Sending HTTP request {HttpMethod} {Uri}";
+            private static readonly string MessageRequestEnd = "Recieved HTTP response {HttpMethod} {Uri} - {StatusCode}";
+
+            static Log()
+            {
+                _requestStart = LoggerMessage.Define<HttpMethod, Uri>(LogLevel.Information, EventIds.RequestStart, MessageRequestStart);
```
anurse: Any reason you're doing this instead of an initializer? Just seems different from our usual patterns.

src/Microsoft.Extensions.Http/Resources.resx
```diff
@@ -117,18 +117,18 @@
...
+  <data name="HttpMessageHandlerBuilder_AdditionalHandlerIsNull" xml:space="preserve">
     <value>The '{0}' must not contain a null entry.</value>
   </data>
-  <data name="DefaultHttpMessageHandlerBuilder_AdditionHandlerIsInvalid" xml:space="preserve">
+  <data name="HttpMessageHandlerBuilder_AdditionHandlerIsInvalid" xml:space="preserve">
     <value>The '{0}' property must be null. '{1}' instances provided to '{2}' must not be reused or cached.{3}Handler: '{4}'</value>
     <comment>0 = nameof(DelegatingHandler.InnerHandler)
 1 = nameof(DelegatingHandler)
-2 = nameof(AdditionalHandlers)
+2 = nameof(HttpMessageHandlerBuilder)
```
anurse: This makes me wish we had a RESX generator that supported syntax like this (including naming parameters) ;)

samples/HttpClientFactorySample/Program.cs
rynowak: Yeah, I think we will need the builder.

src/Microsoft.Extensions.Http/DefaultHttpMessageHandlerBuilderPostInitializer.cs
rynowak: What's the suggestion here exactly?

src/Microsoft.Extensions.Http/Logging/LoggingHttpMessageHandler.cs
rynowak: > Is that because the extension methods were public? Would making them internal just solve that?

No, the problem is about other components inside the assembly.

src/Microsoft.Extensions.Http/Logging/EventIds.cs
rynowak: These eventIds are shared across multiple classes writing to the same logger category.

src/Microsoft.Extensions.Http/Logging/EventIds.cs
rynowak: Ok, I'll just not do this.

src/Microsoft.Extensions.Http/Logging/EventIds.cs
rynowak: I don't want to use nameof here. These names should **not** get refactored.

src/Microsoft.Extensions.Http/DefaultHttpMessageHandlerBuilderPostInitializer.cs
rynowak: > This feels exactly like an IStartupFilter to me. Why wouldn't we follow that pattern? Is there anything fundamental about it that wouldn't work? 

This would work fine. I started with something like that, but I didn't want the complication of having multiple unless we think it's important. 

> Would we want to support more than one though? Is this a "real" extension point or just a hack we're doing to ensure we can put stuff in that wraps the pipeline?

I don't like putting the logging dependency right in the builder. It makes the hairs on the back of my neck stand up. I think if it's useful for others then it's not bad. 

>And, if we're hard coding this in our core package, maybe we should just go the full distance and always do the logging stuff.

I think it's important that it be removable. Some guy will really want that. If you have to replace a lot of default stuff to remove it, then I'd agree with *some guy*.

src/Microsoft.Extensions.Http/Logging/LoggingHttpMessageHandler.cs
rynowak: Oh you mean field initializers. Sure

rynowak: Oh I remember why, readability. But these are going to be an unreadable mess no matter what so #YOLO

rynowak:
It sounds like the salient comments here are:
- https://github.com/aspnet/HttpClientFactory/pull/22#discussion_r148717074
- https://github.com/aspnet/HttpClientFactory/pull/22#discussion_r148716641

Regarding the `IStartupFilter` approach, it seems like a big hammer for this, but I'm not sure that's a real objection. I'm going to go ahead and code that up so we can look at it.

Regarding the 'IHttpClientBuilder' I'm going to go ahead and code that up as an extra commit.

src/Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs
davidfowl: I dislike polluting the top level IServiceCollection with features of subcomponents. I'd need to see a few extension methods before I judge though. I think having a more specific type is better for scoping.

rynowak:
@anurse @davidfowl updated

---

If either of you feel like there are more major design issues here that aren't blocking, I'd like to spin those off into follow ups.

rynowak:
@anurse any more important feedback? Hoping this is ready to merge

PR #24, opened 2017-11-07
Set Description on Microsoft.Extensions.Http

natemcmaster:
Required to unblock moving forward to the 2.1. NuGet [used to set this property](https://github.com/NuGet/NuGet.Client/pull/1648) by default in all projects, causing a false-positive test pass on from the NuGetPackageVerifier. They fixed this...which means our build will actually fail when this isn't set.

cref https://github.com/aspnet/BuildTools/pull/457


rynowak:
:shipit: for now to unblcok

@Eilon @glennc do you want to craft one of these for realsies or want me to the do?

Eilon:
I vote for @rynowak + @glennc coming up with a real description. I logged https://github.com/aspnet/HttpClientFactory/issues/25 to track.

PR #31, opened 2017-11-13
A rough cut at typed clients

rynowak:
This PR has **infrastructure** for typed clients. You either need to
write your own client classes, or use a library like refit.

What's here is the general pattern and plumbing. The intent is that a
library like refit would implement ITypedHttpClientFactory, OR you would
plug it using the overloads of AddTypedClient<T> that take a factory
function.

I'm introducing this now to try and get the discussion started about how
you might use this in MVC.


src/Microsoft.Extensions.Http/ITypedHttpClientFactory.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Net.Http;
+
+namespace Microsoft.Extensions.Http
+{
+    public interface ITypedHttpClientFactory
```
rynowak: A library like Refit can implement this and **just work** with our extension methods.

rynowak:
@davidfowl - you have been summoned.

src/Microsoft.Extensions.Http/DependencyInjection/HttpClientBuilderExtensions.cs
```diff
@@ -80,10 +80,78 @@ public static IHttpClientBuilder AddHttpMessageHandlerBuilderOptions(this IHttpC
...
+
+        public static IHttpClientBuilder AddTypedClient<T>(this IHttpClientBuilder builder)
+            where T : class
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            builder.Services.AddTransient<T>(s =>
```
davidfowl: nit: Does it need to be all service locatory?

rynowak: We could get it down to one call instead of two, but it would require more codes. Factory functions always lead to service resolvers when they are non-trivial

rynowak:
I've been working on this with @glennc during the week at SpringOne #holla

I'm going to go ahead and merge when the CI is green since everyone was happy with the direction overall.

PR #33, opened 2017-11-14
Add Rider to .gitignore

JosephWoodward:


rynowak:
Sorry for the delay in getting to this, I would have merged right away but I didn't see it 😢

JosephWoodward:
No rush, just something I noticed whilst browsing through the source locally.

Issue #26, opened 2017-11-10
Eventing and EventCounters for HTTPClientFactory


glennc:
Plans for how we do the right things.

PR #46, opened 2018-01-18
Add handler rotation and expiry

rynowak:
This adds a simple approach to handler rotation and handler expiry that
should have virtually no impact on how users consume HttpClient.

The idea is that handlers sit in an 'active' pool until they expire
(with a configuable fixed lifetime). While they are in the 'active' pool
they can be shared amongst as many HttpClient instances as we need to
create.

When a handler that supports expiry is created, a timer will be queued
to atomically remove it from the pool. A new one will then be created to
take its place.

----

Ultimately we need to dispose the handler that we removed from the
active pool, but it's not safe to do that right away because there are
probably existing leases. Ultimately this is a task for the GC.

When we create the handler, the one that we give to HttpClients is an
instance of a type we control, that is safe to GC without the need for
disposal.

Once we move the handler to the 'expiry' pool, we can downgrade our
reference to the tracking handler to a weak reference. Once we see that
its GCed, it's safe to dispose the inner handler.

----

We ultimately process all of these disposal jobs on a fixed timer that
processes all of the 'expired' items.

Since all of the data flow is unidirectional there isn't much threading
heavy-lifting.


rynowak:
Not included here:
- tests
- updates to the builder API

rynowak:
> When we create the handler, the one that we give to HttpClients is an instance of a type we control, that is safe to GC without the need for disposal.

I also considered the approach where we keep a weakreference for each HttpClient we create, but I didn't like how the math works out in high throughput scenarios. Each 'expired' item could have thousands or millions or weakreferences associated with it. 

This approach is a little sneaky, but allows the GC to do all of the work for us. It's basically like using a finalizer in the end, but without the gotchas of finalizers. It's more deterministic, because we can control how often we clean up.

src/Microsoft.Extensions.Http/ActiveHandlerTrackingEntry.cs
```diff
@@ -0,0 +1,72 @@
...
+            {
+                return; // never expires.
+            }
+            
+            if (Volatile.Read(ref _timerInitialized))
+            {
+                return;
+            }
+
+            StartExpiryTimerSlow(callback);
```
rynowak: This is basically an implementation of `LazyInitializer.EnsureInitialized`.

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
```diff
@@ -87,16 +144,154 @@ internal HttpMessageHandler CreateHandler(string name)
...
+        }
+
+        // Internal so it can be overridden in tests
+        internal virtual void StartCleanupTimer()
+        {
+            lock (_cleanupTimerLock)
+            {
+                if (_cleanupTimer == null)
+                {
+                    _cleanupTimer = new Timer(_cleanupCallback, null, DefaultCleanupInterval, Timeout.InfiniteTimeSpan);
```
rynowak: note: I prefer this pattern with timers - basically setting an alarm that will go off once at some point in the future. Since there's nothing recurring, you don't have any state to dispose. You just clean up each timer as it goes off.

The BCL already optimizes for using lots of short-lived timers.

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
```diff
@@ -87,16 +144,154 @@ internal HttpMessageHandler CreateHandler(string name)
...
+                _cleanupTimer = null;
+            }
+
+            var stopwatch = ValueStopwatch.StartNew();
+
+            var initialCount = _expiredHandlers.Count;
+            Log.CleanupCycleStart(_logger, initialCount);
+
+            var disposedCount = 0;
+            for (var i = 0; i < initialCount; i++)
```
rynowak: There's actually a subtle issue here, we need to prevent cleanup cycles from running concurrently if the time the cycle takes is longer than the period. 

We really don't intend for multiple threads to do cleanup at once. I could make it supported, but it seems better to prevent it.

Will fix in the next iteration.

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
```diff
@@ -87,16 +144,154 @@ internal HttpMessageHandler CreateHandler(string name)
...
+            Log.CleanupCycleEnd(_logger, stopwatch.GetElapsedTime(), disposedCount, remainingCount);
+
+            // We didn't totally empty the cleanup queue, try again later.
+            if (remainingCount > 0)
+            {
+                StartCleanupTimer();
+            }
+        }
+
+        private static class Log
```
NTaylorMullen: I assume you'll pull this out in the non-design version.

rynowak: No this is the pattern we're using. It's much cleaner than a single class with infinity methods

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
```diff
@@ -87,16 +144,154 @@ internal HttpMessageHandler CreateHandler(string name)
...
                     options.HttpMessageHandlerBuilderActions[i](b);
                 }
             }
         }
+
+        // Internal for tests
+        internal void ExpiryTimer_Tick(object state)
+        {
+            var active = (ActiveHandlerTrackingEntry)state;
+            active.DisposeTimer();
```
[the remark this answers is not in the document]
rynowak: How would the layering of this work?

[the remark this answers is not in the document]
NTaylorMullen: Can have a wrapping delegate in your` ActiveHandlerTrackingEntry` which is responsible for calling into timercallback code.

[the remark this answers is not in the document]
rynowak: I chose to address this for now. It results in an additional delegate per handler entry. We might undo this later.

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
```diff
@@ -87,16 +150,195 @@ internal HttpMessageHandler CreateHandler(string name)
...
+            // Wrap the handler so we can ensure the inner handler outlives the outer handler.
+            var handler = new LifetimeTrackingHttpMessageHandler(builder.Build());
+
+            // Note that we can't start the timer here. That would introduce a very very subtle race condition
+            // with very short expiry times. We need to wait until we've actually handed out the handler once
+            // to start the timer.
+            // 
+            // Otherwise it would be possible that we start the timer here, immediately expire it (very short
+            // timer) and then dispose it without ever creating a client. That would be bad. It's unlikely
+            // this would happen, just we want to be sure.
```
anurse: nit: ~just~ but

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
```diff
@@ -87,16 +150,195 @@ internal HttpMessageHandler CreateHandler(string name)
...
+        // Internal for tests
+        internal void ExpiryTimer_Tick(object state)
+        {
+            var active = (ActiveHandlerTrackingEntry)state;
+
+            // The timer callback should be the only one removing from the active collection. If we can't find
+            // our entry in the collection, then this is a bug.
+            var removed = _activeHandlers.TryRemove(active.Name, out var found);
+            Debug.Assert(removed);
+            Debug.Assert(object.ReferenceEquals(active, found.Value));
```
anurse: nit: Messages in the asserts?

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
```diff
@@ -87,16 +150,195 @@ internal HttpMessageHandler CreateHandler(string name)
...
+            StartCleanupTimer();
+        }
+
+        // Internal so it can be overridden in tests
+        internal virtual void StartHandlerEntryTimer(ActiveHandlerTrackingEntry entry)
+        {
+            entry.StartExpiryTimer(ExpiryTimer_Tick);
+        }
+
+        // Internal so it can be overridden in tests
```
anurse: `private protected`? #usethemnewfeatures

anurse: FYI, it does work with `InternalsVisibleTo`, to my knowledge.

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
anurse: Good to know. Might want to do something similar with timeouts in SignalR :)

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
```diff
@@ -87,16 +150,195 @@ internal HttpMessageHandler CreateHandler(string name)
...
+                Log.CleanupCycleStart(_logger, initialCount);
+
+                var stopwatch = ValueStopwatch.StartNew();
+
+                var disposedCount = 0;
+                for (var i = 0; i < initialCount; i++)
+                {
+                    // Since we're the only one removing from _expired, TryDequeue must always succeed.
+                    _expiredHandlers.TryDequeue(out var entry);
+                    Debug.Assert(entry != null);
```
anurse: nit: message on assert

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
anurse: @BrennanConroy we should do this in SignalR.

src/Microsoft.Extensions.Http/Logging/EventIds.cs
```diff
@@ -23,5 +23,10 @@ internal static class EventIds
         public static readonly EventId ClientHandlerRequestHeader = new EventId(102, "RequestHeader");
         public static readonly EventId ClientHandlerResponseHeader = new EventId(103, "ResponseHeader");
 
+        // Logging for the DefaultHttpClientFactory
+        public static readonly EventId FactoryCleanupCycleStart = new EventId(100, "CleanupCycleStart");
+        public static readonly EventId FactoryCleanupCycleEnd = new EventId(101, "CleanupCycleEnd");
+        public static readonly EventId FactoryCleanupItemFailed = new EventId(102, "CleanupItemFailed");
+        public static readonly EventId FactoryHandlerExpired = new EventId(103, "HandlerExpired");
```
anurse: Hrm, didn't notice this fully in previous reviews. Why is this not part of the `Log` nested class? Aren't we just substituting one large comment-delimited class for another? Also, putting them in `Log` means the names can use `nameof`.

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
BrennanConroy: I know, you made an issue for it, after preview1

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
rynowak: We're not using 7.2 here yet - I'd rather wait until that's turned on globally.

src/Microsoft.Extensions.Http/Logging/EventIds.cs
rynowak: yeah I think you're right about this. I'm not going to use nameof though. I want these to be hardcoded constants because we're not allowed to change them.

rynowak:
@anurse @NTaylorMullen updated

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
```diff
@@ -4,27 +4,66 @@
...
         private readonly IOptionsMonitor<HttpClientFactoryOptions> _optionsMonitor;
         private readonly IHttpMessageHandlerBuilderFilter[] _filters;
+        private readonly Func<string, Lazy<ActiveHandlerTrackingEntry>> _entryFactory;
+
+        // Default time of 30s for cleanup seems reasonable.
+        // Quick math:
+        // 10 distinct named clients * expiry time >= 1s = approximate cleanup queue of 300 items
+        //
+        // This seems frequent enough. We also rely on GC occurring to actually trigger disposal.
+        private readonly TimeSpan DefaultCleanupInterval = TimeSpan.FromSeconds(10);
```
NTaylorMullen: Need to update comment.

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
```diff
@@ -4,27 +4,66 @@
...
+using Microsoft.Extensions.Internal;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.Extensions.Http
 {
     internal class DefaultHttpClientFactory : IHttpClientFactory
     {
+        // establishing a minimum lifetime helps us avoid some possible destructive cases.
+        private readonly TimeSpan MinimumHandlerLifetime = TimeSpan.FromSeconds(1);
```
NTaylorMullen: You don't use this anymore.

test/Microsoft.Extensions.Http.Test/DefaultHttpClientFactoryTest.cs
```diff
@@ -203,5 +219,279 @@ public void Factory_CreateClient_FiltersCanDecorateBuilder()
...
+            var activeEntry2 = Assert.Single(factory._activeHandlers).Value.Value;
+            Assert.Equal("github", activeEntry1.Name);
+            Assert.Equal(TimeSpan.FromMinutes(2), activeEntry1.Lifetime);
+            Assert.NotNull(activeEntry1.Handler);
+            Assert.NotSame(activeEntry1, activeEntry2);
+            Assert.NotSame(activeEntry1.Handler, activeEntry2.Handler);
+        }
+
+        [Fact]
+        public async Task Factory_CreateClient_WithExpiry_CanHandlerCanBeReusedBeforeExpiry()
```
NTaylorMullen: `Factory_CreateClient_WithExpiry_HandlerCanBeReusedBeforeExpiry` (just changing `CanHandler` => `Handler`)

test/Microsoft.Extensions.Http.Test/DefaultHttpClientFactoryTest.cs
```diff
@@ -203,5 +219,279 @@ public void Factory_CreateClient_FiltersCanDecorateBuilder()
...
+            factory.CleanupTimer_Tick(null);
+
+            // Assert
+            Assert.Empty(factory._expiredHandlers);
+            Assert.Equal(1, disposeHandler.DisposeCount);
+            Assert.False(factory.CleanupTimerStarted.IsSet);
+        }
+
+        [Fact]
+        public async Task Factory_CleanupCycle_DoesDisposeLiveHandler()
```
NTaylorMullen: `DoesDispose` => `Disposes`

src/Microsoft.Extensions.Http/Logging/EventIds.cs
```diff
@@ -8,20 +8,15 @@ namespace Microsoft.Extensions.Http.Logging
     internal static class EventIds
```
NTaylorMullen: This is internal and all event ids are nested now. Suggest removing this entirely.

rynowak: Oh oops, this is supposed to be gone

src/Microsoft.Extensions.Http/DependencyInjection/HttpClientBuilderExtensions.cs
```diff
@@ -418,5 +419,45 @@ public static IHttpClientBuilder AddTypedClient<TClient>(this IHttpClientBuilder
...
+        /// </para>
+        /// </remarks>
+        public static IHttpClientBuilder SetHandlerLifetime(this IHttpClientBuilder builder, TimeSpan handlerLifetime)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            if (handlerLifetime < TimeSpan.FromSeconds(1))
```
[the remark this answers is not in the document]
rynowak: I don't super care about parameterizing the resource, That's a very complex thing to solve. I'll define this in a single place and add a comment to update the resource.

src/Microsoft.Extensions.Http/Logging/EventIds.cs
anurse: > I want these to be hardcoded constants because we're not allowed to change them.

Good point.

rynowak:
> Seems like an EventSource and/or EventCounters might be nice for this, to track active handlers and ensure that they're actually getting rotated properly.

Currently planned for preview-2 https://github.com/aspnet/HttpClientFactory/issues/26 


@anurse - any other feedback? I think I've addressed all of the comments so far

PR #52, opened 2018-01-25
Replace hard-coded strings in EventIds class

scottaddie:
Make use of C#'s `nameof` operator to replace some hard-coded string values in the `EventIds` class.


BrennanConroy:
https://github.com/aspnet/HttpClientFactory/pull/46#discussion_r162702246

glennc:
As per linked comment. This is by design.

PR #53, opened 2018-01-25
Allowing Timeout.InfiniteTimeSpan for HandlerLifeTime

### LINKED ISSUE #50: SetHandlerLifetime should allow Timeout.InfiniteTimeSpan
Our `SetHandlerLifetime` method is intended to accept a value of `Timeout.InfiniteTimeSpan` and interpret this as a gesture to disable the handler rotation feature. Unfortunately the validation logic is overzealous.

This is a bug and we need to allow this. We intended for this to work and even documented it 👍

stevejgordon:
Fixes #50

Also updated the setter on HttpClientFactoryOptions.HandlerLifetime

@rynowak I happened to be perusing the code and the latest issues. Thought I'd throw this out there so see if it's what you intended?

The ActiveHandlerTrackingEntry will never expire if the LifeTime is set to InfiniteTimeSpan.


rynowak:
This looks great. Thanks!

Issue #8, opened 2017-10-06
Add Polly


rynowak:
Todo like for this:
- [x] Add basic polly handler and builder support
- [x] Add support for polly context
- [ ] Add logging
- [x] Add builder methods for integration with the policy registry https://github.com/aspnet/HttpClientFactory/pull/58#issuecomment-366183949
- [x] Do we need an opinionated **retry-focused** extension method https://github.com/aspnet/HttpClientFactory/pull/58#discussion_r168934283

For docs:
- [x] Add documentation about timeouts, see: https://github.com/aspnet/HttpClientFactory/pull/58#issuecomment-366283187
- [x] Add documentation about kinds of policies and which builder methods to use https://github.com/aspnet/HttpClientFactory/pull/58#discussion_r168352051

stevejgordon:
Can I check... is the intention to include some pre-designed DelegatingHandlers utilising Polly for retry, circuit breaker, caching etc "in the box" with the release? Or simply to include some samples? I'm considering starting to develop a policy inside a delegating handler of my own for testing an approach to something we currently do more manually. Trying to decide how much time to focus on that and whether to blog about it or wait until I see any default ones.

rynowak:
I think it's important that whatever we provide exposes Polly's API in a high fidelity way. They have a lot of different options and ways of doing things. 

Goals:
1. Expose their APIs (don't create wrappers)
2. Make it possible to use most of what they provide without having to author a handler
3. Provide a good logging experience for common usage

Any time we interact with a community project we're ask the question, "why should **Microsoft** build this?" A good reason for Microsoft to build it in this case is that we think having access to Polly for retries and circuit breakers is pretty fundamental. We also really want the logging to great. 

------


Right now my (and @glennc)'s thinking is that we'll provide an extension on the builder method that looks like `.AddPollyHandler(Func<AsyncPolicyBuilder, AsyncPolicyBuilder>)`. This is far from decided.


If you compare this to [my sample](https://github.com/rynowak/HttpClientFactoryDemos/blob/master/PollySample/RetryHandler.cs#L19) - it means that you would write code like:

```
.AddPollyHandler(policy.WaitAndRetryAsync(5, (n) => TimeSpan.FromSeconds(n))
```

and the result would be what I showed in my sample.

There are a few challenges with this, naming figuring out how we get logging into the picture.

stevejgordon:
@rynowak Makes sense 

[...]

PR #58, opened 2018-02-14
Start to flesh out Polly design

rynowak:
The makings of some integration for using HttpClientFactory and Polly together.

Right now my focus is on nailing down the **basics** of the public-facing API:
- What are the most common scenarios?
- How does a beginner figure out what to do?
- Does an expert have enough control?

My goals is to answer those questions (with feedback) along with the the following non-goals:
- Don't lock users out of powerful but complicated things
- Don't try to address every possible scenario and combination
- Don't hide or abstract away Polly's APIs

### DIFF
--- a/samples/HttpClientFactorySample/Program.cs
+++ b/samples/HttpClientFactorySample/Program.cs
@@ -8,6 +8,7 @@ using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json.Linq;
+using Polly;
 
 namespace HttpClientFactorySample
 {
@@ -53,6 +54,13 @@ namespace HttpClientFactorySample
                 c.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json"); // Github API versioning
                 c.DefaultRequestHeaders.Add("User-Agent", "HttpClientFactory-Sample"); // Github requires a user-agent
             })
+
+            // Build a totally custom policy using any criteria
+            .AddPolicyHandler(Policy.Handle<HttpRequestException>().RetryAsync())
+
+            // Build a policy that will handle exceptions and 500s from the remote server
+            .AddServerErrorPolicyHandler(p => p.RetryAsync())
+
             .AddHttpMessageHandler(() => new RetryHandler()) // Retry requests to github using our retry handler
             .AddTypedClient<GithubClient>();
         }
--- a/src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
+++ b/src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
@@ -1,0 +1,125 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Net;
+using System.Net.Http;
+using Microsoft.Extensions.Http;
+using Polly;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    /// <summary>
+    /// Extensions methods for configuring <see cref="PolicyHttpMessageHandler"/> message handlers as part of
+    /// and <see cref="HttpClient"/> message handler pipeline.
+    /// </summary>
+    public static class PollyHttpClientBuilderExtensions
+    {
+        /// <summary>
+        /// Adds a <see cref="PolicyHttpMessageHandler"/> which will surround request execution with the provided
+        /// <see cref="IAsyncPolicy"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="IHttpClientBuilder"/>.</param>
+        /// <param name="policy">The <see cref="IAsyncPolicy"/>.</param>
+        /// <returns>An <see cref="IHttpClientBuilder"/> that can be used to configure the client.</returns>
+        /// <remarks>
+        /// <para>
+        /// See the remarks on <see cref="PolicyHttpMessageHandler"/> for guidance on configuring policies.
+        /// </para>
+        /// </remarks>
+        public static IHttpClientBuilder AddPolicyHandler(this IHttpClientBuilder builder, IAsyncPolicy policy)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            if (policy == null)
+            {
+                throw new ArgumentNullException(nameof(policy));
+            }
+
+            // Important - cache policy instances so that they are singletons per handler.
+            var innerPolicy = policy.WrapAsync(Policy.NoOpAsync<HttpResponseMessage>());
+
+            builder.AddHttpMessageHandler(() => new PolicyHttpMessageHandler(innerPolicy));
+            return builder;
+        }
+
+        /// <summary>
+        /// Adds a <see cref="PolicyHttpMessageHandler"/> which will surround request execution with the provided
+        /// <see cref="IAsyncPolicy{HttpResponseMessage}"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="IHttpClientBuilder"/>.</param>
+        /// <param name="policy">The <see cref="IAsyncPolicy{HttpResponseMessage}"/>.</param>
+        /// <returns>An <see cref="IHttpClientBuilder"/> that can be used to configure the client.</returns>
+        /// <remarks>
+        /// <para>
+        /// See the remarks on <see cref="PolicyHttpMessageHandler"/> for guidance on configuring policies.
+        /// </para>
+        /// </remarks>
+        public static IHttpClientBuilder AddPolicyHandler(this IHttpClientBuilder builder, IAsyncPolicy<HttpResponseMessage> policy)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            if (policy == null)
+            {
+                throw new ArgumentNullException(nameof(policy));
+            }
+
+            builder.AddHttpMessageHandler(() => new PolicyHttpMessageHandler(policy));
+            return builder;
+        }
+
+        /// <summary>
+        /// Adds a <see cref="PolicyHttpMessageHandler"/> which will surround request execution with a <see cref="Policy"/>
+        /// created by executing the provided configuration delegate. The policy builder will be preconfigured to trigger
+        /// application of the policy for requests that fail with a connection or server error (5XX status code).
+        /// </summary>
+        /// <param name="builder">The <see cref="IHttpClientBuilder"/>.</param>
+        /// <param name="configurePolicy">A delegate used to create a <see cref="IAsyncPolicy{HttpResponseMessage}"/>.</param>
+        /// <returns>An <see cref="IHttpClientBuilder"/> that can be used to configure the client.</returns>
+        /// <remarks>
+        /// <para>
+        /// See the remarks on <see cref="PolicyHttpMessageHandler"/> for guidance on configuring policies.
+        /// </para>
+        /// <para>
+        /// The <see cref="PolicyBuilder{HttpResponseMessage}"/> provided to <paramref name="configurePolicy"/> has been
+        /// preconfigured to handle connection errors (as <see cref="HttpRequestException"/>) or server errors (as a 5XX HTTP
+        /// status code). The configuration is similar to the following code sample:
+        /// <code>
+        /// Policy.HandleAsync&lt;HttpRequestException&gt;().OrResult&lt;HttpResponseMessage&gt;(response =>
+        /// {
+        ///     return response.StatusCode >= HttpStatusCode.InternalServerError;
+        /// }
+        /// </code>
+        /// </para>
+        /// <para>
+        /// The policy created by <paramref name="configurePolicy"/> will be cached indefinitely per named client. Policies
+        /// are generally designed to act as singletons, and can be shared when appropriate. To share a policy across multiple
+        /// named clients, first create the policy and the pass it to multiple calls to 
+        /// <see cref="AddPolicyHandler(IHttpClientBuilder, IAsyncPolicy)"/> or
+        /// <see cref="AddPolicyHandler(IHttpClientBuilder, IAsyncPolicy{HttpResponseMessage})"/> as desired.
+        /// </para>
+        /// </remarks>
+        public static IHttpClientBuilder AddServerErrorPolicyHandler(
+            this IHttpClientBuilder builder, 
+            Func<PolicyBuilder<HttpResponseMessage>, IAsyncPolicy<HttpResponseMessage>> configurePolicy)
+        {
+            
+            var policyBuilder = Policy.Handle<HttpRequestException>().OrResult<HttpResponseMessage>(response =>
+            {
+                return response.StatusCode >= HttpStatusCode.InternalServerError;
+            });
+
+            // Important - cache policy instances so that they are singletons per handler.
+            var policy = configurePolicy(policyBuilder);
+
+            builder.AddHttpMessageHandler(() => new PolicyHttpMessageHandler(policy));
+            return builder;
+        }
+    }
+}
--- a/src/Microsoft.Extensions.Http.Polly/Microsoft.Extensions.Http.Polly.csproj
+++ b/src/Microsoft.Extensions.Http.Polly/Microsoft.Extensions.Http.Polly.csproj
@@ -1,0 +1,25 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>
+      The HttpClient factory is a pattern for configuring and retrieving named HttpClients in a composable way. This package integrates IHttpClientFactory with the Polly library, to add transient-fault-handling and resiliency through fluent policies such as Retry, Circuit Breaker, Timeout, Bulkhead Isolation, and Fallback.
+    </Description>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageTags>aspnetcore;httpclient</PackageTags>
+
+    <!-- Don't use Microsoft.AspNetCore.Http.Polly as a namespace, that introduces ambiguities with 'Polly' -->
+    <RootNamespace>Microsoft.AspNetCore.Http</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.Extensions.Http\Microsoft.Extensions.Http.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Polly-Signed" Version="$(PollySignedPackageVersion)" />
+    <PackageReference Include="Microsoft.Extensions.ValueStopwatch.Sources" Version="$(MicrosoftExtensionsValueStopwatchSourcesPackageVersion)" PrivateAssets="All" />
+  </ItemGroup>
+
+</Project>
--- a/src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
+++ b/src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
@@ -1,0 +1,83 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Polly;
+
+namespace Microsoft.Extensions.Http
+{
+    /// <summary>
+    /// A <see cref="DelegatingHandler"/> implementation that executes request processing surrounded by a <see cref="Policy"/>.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// This message handler implementation supports the use of policies provided by the Polly library for 
+    /// transient-fault-handling and resiliency.
+    /// </para>
+    /// <para>
+    /// The documentation provided here is focused guidance for using Polly together with the <see cref="IHttpClientFactory"/>. 
+    /// See the Polly project and its documentation (https://github.com/app-vnext/Polly) for authoritative information on Polly.
+    /// </para>
+    /// <para>
+    /// The extension methods on <see cref="PollyHttpClientBuilderExtensions"/> are designed as a convenient and correct
+    /// to creating a <see cref="PolicyHttpMessageHandler"/>.
+    /// </para>
+    /// <para>
+    /// The <see cref="PollyHttpClientBuilderExtensions.AddPolicyHandler(IHttpClientBuilder, IAsyncPolicy)"/> and 
+    /// <see cref="PollyHttpClientBuilderExtensions.AddPolicyHandler(IHttpClientBuilder, IAsyncPolicy{HttpResponseMessage})"/>
+    /// methods support the creation of a <see cref="PolicyHttpMessageHandler"/> for any kind of policy. This includes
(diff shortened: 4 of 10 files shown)


rynowak:
Pinging everyone who's commented on the issue thread, I welcome your thoughts and feedback. @stevejgordon @theezak @reisenberger and @joelhulen as well as @glennc @javiercn  and @anurse for good measure

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
```diff
@@ -0,0 +1,36 @@
...
+
+using Polly;
+using System;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.Http
+{
+    public class PolicyHttpMessageHandler : DelegatingHandler
```
rynowak: The key building block here is a message handler that plugs in a policy to the pipeline. Ideally this is as generic as possible.

Not seen here: we'll want logging as well as that's one of the best ways that us building the integration adds value. I'd like to defer discussions about that until after the initial API for configuring clients is designed.

samples/HttpClientFactorySample/Program.cs
```diff
@@ -53,6 +54,19 @@ public static void Configure(IServiceCollection services)
...
+            .AddPolicyHandler(Policy.Handle<HttpRequestException>().RetryAsync())
+
+            // Build a policy that will handle exceptions (connection failures)
+            .AddExceptionPolicyHandler(p => p.RetryAsync())
+
+            // Build a policy that will handle exceptions and 500s from the remote server
+            .AddServerErrorPolicyHandler(p => p.RetryAsync())
+
+            // Build a policy that will handle exceptions, 400s, and 500s from the remote server
+            .AddBadRequestPolicyHandler(p => p.RetryAsync())
```
rynowak: This is probably the best place to start the discussion. 

I understand the requirements for building a message handler to be:
- what kinds of things do I consider an error?
- what do I want to do when I find one?

I think there's some value in us providing guidance about the first question (what should I consider an error), and I think the second question is easy enough for users to answer (`p => p.RetryAsync()`).

I've defined a few flavors of extension methods here. I think `AddServerErrorPolicyHandler` is probably going to be the most useful balance between "sane default" and hiding complexity. Most folks will always want to treat a connection failure or a 500 as an error.

If you use `AddPolicyHandler` to build `AddServerErrorPolicyHandler` it will end up looking like:
```
.AddPolicyHandler(
  Policy
    .Handle<HttpRequestException>()
    .OrResult<HttpResponseMessage>(r => r.StatusCode >= HttpStatusCode.InternalServerError)
    .RetryAsync());
```
I think it's a good enough for us to find an attractive name for this and include it 👍 

A few questions for discussion:
- Handlers may be long-lived or short-lived. Is it important we reuse policies? It is important that we don't?
- There's a little bit of tension around whether or not things should be `IAsyncPolicy` or `IAsyncPolicy<T>` in some of the simple cases, so I planned to support both. Is this a good plan?

src/Microsoft.Extensions.Http.Polly/Microsoft.Extensions.Http.Polly.csproj
```diff
@@ -0,0 +1,25 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>
+      The HttpClient factory is a pattern for configuring and retrieving named HttpClients in a composable way. This package provides integration for IHttpClientFactory using the Polly library for exception handling and fault-tolerance.
```
rynowak: @reisenberger @joelhulen - feel free to provide feedback about how we're describing your project 😆

joelhulen: Not a bad start :)

I'd perhaps word the second sentence as such:

This package integrates IHttpClientFactory with the Polly library, to add transient-fault-handling and resiliency through fluent policies such as Retry, Circuit Breaker, Timeout, Bulkhead Isolation, and Fallback.

@reisenberger, do you agree?

samples/HttpClientFactorySample/Program.cs
Kralizek: I jump in even if not pinged. Policies' lifecycle should be independent from the handler's one because there are policies like the short-circuit one that are designed to be "singleton-ish".

Ideally policies should be attached to freshly generated handler's but it feels like it leads to a pool manager for the policies to be used by a pool manager for the handlers...

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            if (request == null)
+            {
+                throw new ArgumentNullException(nameof(request));
+            }
+
+            return _policy.ExecuteAsync(() => base.SendAsync(request, cancellationToken));
```
martincostello: Polly supports handing the cancellation token to the policy itself as well so it can be used by its internals (at least that's my understanding), which is passed through into the lambda doing the actual work. Should the cancellation token be used that way here?

reisenberger: Agree, eg: 

    return _policy.ExecuteAsync(token => base.SendAsync(request, token), cancellationToken);

An example benefit: if the user has configured a wait-and-retry policy, a `cancellationToken` passed through the policy in this manner also governs the wait phase of wait-and-retry, so that signalling the `cancellationToken` will cause the policy to abandon the wait immediately.

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
reisenberger: :+1: to this highly general `PolicyHttpMessageHandler`

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+            var policyBuilder = Policy.Handle<HttpRequestException>().OrResult<HttpResponseMessage>(response =>
+            {
+                return response.StatusCode >= HttpStatusCode.InternalServerError;
+            });
+            var policy = createPolicy(policyBuilder);
+
+            //builder.AddHttpMessageHandler(() => new PolicyHttpMessageHandler(createPolicy()));
+            return builder;
+        }
+
```
reisenberger: Polly policies divide into reactive policies (Retry; CircuitBreaker; Fallback), and proactive and preemptive policies (Cache; Bulkhead; Timeout).  The reactive policies are the ones which flow from `PolicyBuilder`, so the choice of configuration syntax here with variants around `Func<PolicyBuilder..., IAsyncPolicy...>` focuses on (only) the Retry, CircuitBreaker and Fallback policies.

I'm drawing this out in case it wasn't intended, but it seems reasonable, as Retry and CircuitBreaker seem likely majority use cases for Polly with `HttpClientFactory`.  And for those who want to go further, the more advanced overloads (eg `AddPolicyHandler(this IHttpClientBuilder builder, IAsyncPolicy policy)` line 14) do provide.

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+        public static IHttpClientBuilder AddExceptionPolicyHandler(this IHttpClientBuilder builder, Func<PolicyBuilder, IAsyncPolicy> createPolicy)
+        {
+            var policyBuilder = Policy.Handle<HttpRequestException>();
+            var policy = createPolicy(policyBuilder);
+
+            //builder.AddHttpMessageHandler(() => new PolicyHttpMessageHandler(createPolicy()));
+            return builder;
+        }
+
+        public static IHttpClientBuilder AddExceptionPolicyHandler(this IHttpClientBuilder builder, Func<PolicyBuilder, IAsyncPolicy<HttpResponseMessage>> createPolicy)
```
reisenberger: Point of detail: this overload is not needed.  Polly fluent syntax happens to dictate that `PolicyBuilder` must already be generic `PolicyBuilder<TResult>` if the eventual policy is to be generic in `TResult`. (no configuration overloads flow from `PolicyBuilder` to `IAsyncPolicy<TResult>`.)

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+            var policyBuilder = Policy.Handle<HttpRequestException>().OrResult<HttpResponseMessage>(response =>
+            {
+                return response.StatusCode >= HttpStatusCode.BadRequest;
+            });
+            var policy = createPolicy(policyBuilder);
+
+            //builder.AddHttpMessageHandler(() => new PolicyHttpMessageHandler(createPolicy()));
+            return builder;
+        }
+    }
```
reisenberger: Is there as clear motivation for an `AddBadRequestPolicyHandler()` covering the 4xx codes generally, as for the other overloads?  Where Http status codes in the 4xx range indicate client errors, there may be no point in retrying these (eg if a request is badly formed 400, resubmitting it is usually not going to make it succeed).  Likewise, one typically might not want to take client-errors as cause to break a circuit-breaker.

Some specific exceptions: 

+ 408 (afaiu) the server signalling giving up waiting for the client's request to arrive complete; it could reflect wider network connectivity which merits a retry.
+ 429 RetryAfter: Polly has retry overloads which can handle this case (extract how long to wait before the next try, from the 429 response), but it remains a fairly specific configuration.

stevejgordon:
I like where this is headed :+1:

stevejgordon:
I was playing with this briefly last night and thinking about the scenario we'd likely have were we like to register and reuse policies. This is what I ended up with...

_NOTE: It's a bit rushed and I didn't have time to / couldn't actually get the sample running so this may very well not work in practice!_

Add an extension:

```csharp
public static IHttpClientBuilder AddPolicyHandler<T>(this IHttpClientBuilder builder, string namedPolicy) where T : IPolicyRegistry<string>
{
	IReadOnlyPolicyRegistry<string> registry;

	using (var sp = builder.Services.BuildServiceProvider())
	{
		registry = sp.GetService<T>();
	}
	
	if (registry != null)
	{
		if (registry.TryGet<IAsyncPolicy<HttpResponseMessage>>(namedPolicy, out var policy))
		{
			builder.AddHttpMessageHandler(() => new PolicyHttpMessageHandler(policy));
		}
	}

	return builder;
}
```

Basic consumption...

```csharp
public static void Configure(IServiceCollection services)
{
	services.AddSingleton<PolicyRegistry>(s =>
	{
		var registry = new PolicyRegistry();

		var policy = Policy.Handle<HttpRequestException>().RetryAsync();

		registry.Add("mypolicy", policy);

		return registry;
	});

	services.AddHttpClient("github", c =>
	{
		c.BaseAddress = new Uri("https://api.github.com/");

		c.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json"); // Github API versioning
		c.DefaultRequestHeaders.Add("User-Agent",
			"HttpClientFactory-Sample"); // Github requires a user-agent
	})

	.AddPolicyHandler<PolicyRegistry>("mypolicy");
}
```

I'm not sure if this is a) a good approach b) useful as part of the main library but interested in thoughts on this style of usage!

rynowak:
Thanks for all of the comments, I'm super excited to see so much interest. I'm going to go through and respond to everyone's feedback, and I'll create a more complete version of this PR over the weekend.

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
rynowak: @reisenberger - do you think that we'll end up having a generic and non-generic policy handler?  That seems logical to me based on the desire to provide both:
```
public static IHttpClientBuilder AddPolicyHandler(this IHttpClientBuilder builder, IAsyncPolicy policy)
public static IHttpClientBuilder AddPolicyHandler(this IHttpClientBuilder builder, IAsyncPolicy<HttpResponseMessage> policy)
```

Do you agree with this direction?

samples/HttpClientFactorySample/Program.cs
rynowak: @Kralizek thanks!

The factory already provides some pooling and lifecycle management for handlers (docs in progress). 

- It's easy to make a policy a singleton. 
- It's also easy to make a policy have the same lifetime as a handler. 
- It would be a bit tricky to do something different, and I'd like to avoid it if possible

> Policies' lifecycle should be independent from the handler's one because there are policies like the short-circuit one that are designed to be "singleton-ish".

This makes intuitive sense to me. 

Are there real use cases for making policies that have the same lifetime as the handler? Should they all be singleton?

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
rynowak: Yep, this is just something I missed. Will fix in the next iteration.

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
rynowak: Thanks for the feedback. This will be good info for writing the docs.

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
rynowak: Great! will remove this in the next iteration.

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
rynowak: > Is there as clear motivation for an AddBadRequestPolicyHandler() covering the 4xx codes generally, as for the other overloads?

Not really. That was just a straw man for us to beat up. 

I still want to get more feedback on whether we should have both `AddExceptionPolicyHandler` and `AddServerErrorPolicyHandler`.. When might you want to apply a policy to a connection failure but not a 500? Still needs more thought IMO.

rynowak:
Responding to a comment from @aristovg

> Sorry for jumping in, but have you considered the HttpClient timeout when doing retries within delegating handler? Consider the following scenario:

>HttpClient timeout is set to 10 seconds
>Retry policy has 5 attempts
>Each attempts takes 5 seconds (either it is set by retry policy or server is responding at that rate).
>This will result in only 2 retries being made instead of 5.
>Any thoughts on that?


For reference the default timeout is 100 seconds. 

I guess I'd say that with a scheme like this you can configure the overall timeout of the system (including retries) using the timeout on `HttpClient`. You can configure the timeout on the actual HTTP requests by using another handler (after the policy) with a stricter timeout enforced by a cancellation token. 

Polly has policies for timeout as well, so I'd assume that once I address [this feedback](https://github.com/aspnet/HttpClientFactory/pull/58/commits/ad8554ace8957a371523dad8395a220d4aec3f26#r168108921) that would 'just work'.

It's a good point and I'm glad that you brought it up. We will need to document this.

rynowak:
Responding to a comment from @stevejgordon 

> I was playing with this briefly last night and thinking about the scenario we'd likely have were we like to register and reuse policies. This is what I ended up with...

Yeah, I like the direction of what you came up with. This seems promising. I'm tracking this for follow up here: https://github.com/aspnet/HttpClientFactory/issues/8

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
reisenberger: Yes.

While only pure `IAsyncPolicy<HttpResponseMessage>` policies would be needed to govern `SendAsync(...)`, common Polly syntax (for example `Policy.TimeoutAsync(10)`) results more naturally in a non-generic `IAsyncPolicy`, so not supporting `IAsyncPolicy` might frustrate users.  

Generic policies are marginally more performant but any difference is in the order of microseconds or less.

reisenberger:
Responding to @aristovg and @rynowak re timeouts:

>>Sorry for jumping in, but have you considered the HttpClient timeout when doing retries within 
>>delegating handler? Consider the following scenario:

>>HttpClient timeout is set to 10 seconds
>>Retry policy has 5 attempts
>>Each attempts takes 5 seconds (either it is set by retry policy or server is responding at that rate).
>>This will result in only 2 retries being made instead of 5.
>>Any thoughts on that?

>[...] You can configure the timeout on the actual HTTP requests by using another handler (after the 
>policy) with a stricter timeout enforced by a cancellation token.
>Polly has policies for timeout as well [...] so I'd assume [...] that would 'just work'.

Exactly so.  Polly's [TimeoutPolicy](https://github.com/App-vNext/Polly/wiki/Timeout) functions [by this same mechanism](https://github.com/App-vNext/Polly/blob/2862be4f4db5fd2fe4a97e9c55faf4b49793e95d/src/Polly.Shared/Timeout/TimeoutEngineAsync.cs#L24-L33), so one could:

```csharp
.AddServerErrorPolicyHandler(p => p.RetryAsync(/* detailed  configuration */))
.AddPolicyHandler(Policy.TimeoutAsync(10)) // impose 10-second timeout on each individual retry
```

---

Additionally, rather than only letting the `TaskCanceledException` or `OperationCanceledException` propagate, Polly's `TimeoutPolicy` [throws a wrapping `Polly.Timeout.TimeoutRejectedException`, if specifically distinguishing timeout to be the cause](https://github.com/App-vNext/Polly/blob/2862be4f4db5fd2fe4a97e9c55faf4b49793e95d/src/Polly.Shared/Timeout/TimeoutEngineAsync.cs#L54-L59).  This helps any wrapping RetryPolicy distinguish timeout cancellation from other cancellation: for instance, one commonly wants to retry on timeout but bail out immediately (not retry) on user cancellation. 

---

Great to have drawn this point out and agree that the role of each timeout merits documentation.

samples/HttpClientFactorySample/Program.cs
reisenberger: Responding to @rynowak and @Kralizek on policy lifecycle:

**TL;DR** Policies should be single-instance for each policy configured for use with a given named `HttpClient` configuration.  

For example, if a Polly policy for a given named `HttpClient` is config'd by supplying a `createPolicy()` factory Func, that `createPolicy()` func should be used to create one instance of that policy, and that single policy instance supplied to all subsequently created handlers/`HttpClient` instances of that named configuration. 

---

#### Explanation

Two policy types maintain state across calls for their designed function: **circuit-breakers** (tracking call faults/successes) and **bulkheads** (limiting parallelism).  

If `HttpClientFactory` were to create fresh instances when recycling handlers, that state would suffer either discontinuity (if new instances replaced old instances 'instantaneously') or fragmentation/multiplicity (if new and old instances co-existed).

eg **Bulkhead:** Generating fresh BulkheadPolicy instances for new handlers could cause executions to exceed the parallelism configured by the developer, as double/multiple capacity would be available while both newly-spun-up and existing / being-retired instances co-existed.  

No other policies maintain state, but all Polly policies are entirely thread-safe and memory-efficient for long lifetime, so there's no real case for shorter lifetimes; it'd just be more allocations and GCs.

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
reisenberger: >[should we] have both AddExceptionPolicyHandler and AddServerErrorPolicyHandler?
>When might you want to apply a policy to a connection failure but not a 500? 

If the questions asked by users to the Polly team are a relevant measure, then handling them differently seems at least uncommon.  We received a lot of requests for policies to handle result codes as well as exceptions, but since adding that, no requests for help how to handle them differently.  My feeling (fwiw) is the majority use case defines a common strategy for all kinds of transient fault at an endpoint.  

If `AddExceptionPolicyHandler` and `AddServerErrorPolicyHandler` were merged, the power overloads you are proposing would still provide, for those who want more nuanced strategies.

Still interested what others may think.

---

The two enquiries the Polly team handles _periodically_ (I would not say frequently) on behaviour for specific result codes are:

+ 401 Unauthorized: [Can I make a policy to re-authenticate, then retry, on 401 Unauthorised?](https://github.com/App-vNext/Polly/wiki/Retry#retry-to-refresh-authorization)
+ 429 RetryAfter: [How do I make a wait-and-retry policy wait for a duration specified in an `x-ms-retry-after-ms` return header?](https://github.com/App-vNext/Polly/wiki/Retry#retryafter-when-the-response-specifies-how-long-to-wait)

We also in the Polly doco [recommend retrying on 408s](https://github.com/App-vNext/Polly/#step-1b-optionally-specify-return-results-you-want-to-handle), similar to [advice from the Azure team](https://docs.microsoft.com/en-us/azure/architecture/best-practices/retry-service-specific#general-rest-and-retry-guidelines).

rynowak: I think it's worth considering if we want to follow up on a convenience method for 'retryable' faults. Adding it to my todo list.

Regarding AuthZ, we plan to do that in the future (not this release). I'm not sure that plan is related to Polly, but we're interested in your thoughts as well.

rynowak:
Updated this. I think I've addressed all of the important feedback that isn't tracked somewhere else.

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
```diff
@@ -0,0 +1,125 @@
...
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            if (policy == null)
+            {
+                throw new ArgumentNullException(nameof(policy));
+            }
+
+            // Important - cache policy instances so that they are singletons per handler.
+            var innerPolicy = policy.WrapAsync(Policy.NoOpAsync<HttpResponseMessage>());
```
rynowak: @reisenberger - what do you think of this approach instead of creating two separate code paths?

Having two separate code paths just seems a bit like clutter - and then there's naming to think about 😲..

reisenberger: var innerPolicy = policy.WrapAsync(Policy.NoOpAsync<HttpResponseMessage>());

I like how this simplifies the separate codepaths; that concerned me too.

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
reisenberger: >I think it's worth considering if we want to follow up on a convenience method for 'retryable' faults. 

Happy to hear more what is the thinking here.  Is this around helpers for commoner cases of _how_ to handle, or something else?

rynowak: > Happy to hear more what is the thinking here. Is this around helpers for commoner cases of how one might handle (common retry strategies such as exponential backoff), or something else?

I think we'd want to provide a more opinionated approach to *what we consider a 'retry-able' request*. The starting point would be the info that you've already shared with us from your docs/guidance.

I think the polly API for creating different retry policies is already sufficient. 

If you have more ideas about this, feel free to open an issue for more focused discussion. Your feedback has been really valuable.

rynowak:
It sounds like we've pretty much converged at this point, and there are lots of follow up items 👍 

I'm going to go ahead and merge this once I do all the other fancy build stuff I need to do in our private repos.

PR #59, opened 2018-02-14
Change HttpMessageHandler cleanup logs from Debug to Information

lfoust:
After using the library for a while I notice that the debug messages that are constantly being output in the debug window when my app is running make it harder to debug other issues and don't really add much value. If my app is left running for any amount of time in the debugger, any non `HttpClientFactory` debug messages are lost in the noise of cleanup messages. I suggest changing the log messages output during the cleanup cycle from **Debug** to **Information**.

### DIFF
--- a/src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
+++ b/src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
@@ -304,12 +304,12 @@ namespace Microsoft.Extensions.Http
             }
 
             private static readonly Action<ILogger, int, Exception> _cleanupCycleStart = LoggerMessage.Define<int>(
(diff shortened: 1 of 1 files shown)


lfoust:
After trying this in a couple of different dev environments I noticed this solution only works if you filter things a certain way. I am going to remove this PR and open an issue to discuss how best to reduce the logging noise around the HttpMesageHandler cleanup

rynowak:
Yeah, I would expect the VS Debug logger to include Info level logs as well.

Please open an issue and start a discussion about what you're seeing. If you're getting a lot of spam about cleanups then it sounds like something isn't working as intended and we should look into it. We don't expect all of the logging at debug level to be meaningful all of the time, but we also don't expect an unholy amount of spam.

PR #63, opened 2018-02-22
Refine test AddPolicyHandler_Generic_AddsPolicyHandler

reisenberger:
The test was identical to the nongeneric test above it.  From the test title, it looks as if the intention is to exercise the `AddPolicyHandler(this IHttpClientBuilder builder, IAsyncPolicy<HttpResponseMessage> policy)` generic method.  This change makes it do so.


rynowak:
Ahhh oopsies. Thanks

PR #65, opened 2018-02-23
Add support for Context

### LINKED ISSUE #62: Execution-scoped data in the Polly-HttpClientFactory integration: enabling rich and correct telemetry, and CachePolicy
Polly Policy instances can be used across multiple call sites for [intended functional outcomes](https://github.com/aspnet/HttpClientFactory/issues/60#issue-298001498).  This is a good fit for named `HttpClient` configurations on `HttpClientFactory` with a `BaseUri` eg `api.contoso.com` indicating all calls will be to endpoints on `api.contoso.com`.

To complement re-use, Polly provides an execution-scoped [`Context`](https://github.com/App-vNext/Polly/blob/754a4a920900e1ae56d4fdae7cbdfd02bcb8cbd1/src/Polly.Shared/Context.cs) which travels with every execution. This:

(a) provides a [correlation id unique to each execution](https://github.com/App-vNext/Polly/blob/754a4a920900e1ae56d4fdae7cbdfd02bcb8cbd1/src/Polly.Shared/Context.cs#L47-L58), allowing logging and telemetry to correlate the events particular to a single execution;
  
(b) [distinguishes different call sites](https://github.com/App-vNext/Polly/blob/754a4a920900e1ae56d4fdae7cbdfd02bcb8cbd1/src/Polly.Shared/Context.cs#L40-L45) at which Policy instances may be being used, allowing logging and telemetry to correlate statistics particular to a single usage site/call path;

+ For instance, it is typical to use the same circuit-breaker instance for all endpoints on `api.contoso.com`, but when analysing failures, it can be very useful to extract that call path A tended to be healthy but call path B tended to require retries, break circuits, etc.   

(c) allows the functioning of `CachePolicy`.  With `CachePolicy`, one typically configures a single `CachePolicy` instance, then `Context` [supplies the key that the policy should use for caching items](https://github.com/App-vNext/Polly/wiki/Cache#default-cachekeystrategy-if-none-specified) on that particular call path.

The execution-scoped `Polly.Context` is already made available to policy delegate hooks (eg [`onRetry`](https://github.com/App-vNext/Polly/blob/2862be4f4db5fd2fe4a97e9c55faf4b49793e95d/src/Polly.Shared/Retry/RetrySyntaxAsync.cs#L142), [`onBreak`](https://github.com/App-vNext/Polly/blob/2862be4f4db5fd2fe4a97e9c55faf4b49793e95d/src/Polly.Shared/CircuitBreaker/CircuitBreakerSyntaxAsync.cs#L100) etc) and will also be available to forthcoming raw telemetry events.

---

Without HttpClientFactory doing anything around this, correlation ids (a) would be inconsistent in cases where multiple Polly policies are applied by layering DelegatingHandlers (each `PolicyHttpMessageHandler` layer would generate a fresh guid).  And features (b) and (c) would not be available.

#### Options for supporting `Polly.Context` in `PolicyHttpMessageHandler` 

[Here within `PolicyHttpMessageHandler`](https://github.com/aspnet/HttpClientFactory/blob/9b11dd0a31de94acbbfc9e7e06015bc73043f20a/src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs#L32-L33) we could:

```csharp
Polly.Context pollyExecutionContext = request.Properties[PollyContext] ?? new Polly.Context(); // [1]
request.Properties[PollyContext] = pollyExecutionContext; // [2]  

return _policy.ExecuteAsync((context, token) => base.SendAsync(request, token), pollyExecutionContext, cancellationToken); // [3]
```

where:

```csharp
private static readonly string PollyContext = "PollyExecutionContext"; // [4]
```

[1] and [2] ensure when there is a chain of delegating handlers, they all use the same context, fixing (a) above.  [3] just despatches so that all Polly logging and telemetry has access to the Context.

If we changed [4] to:

```csharp
public static readonly string PollyContext = "PollyExecutionContext"; // [4b]
```

then users who desired, would have this usage:

```csharp
request.Properties[PolicyHttpMessageHandler.PollyContext] = new Polly.Context("MyUsageKey"); // [5] 
var response = client.DoWhateverAsync(request...); // (as usual)
```

This enables and fixes both use cases (b) and (c) above.

---

Stepping back: the issue here is that Polly users normally code the `fooPolicy.ExecuteAsync(...)` call directly, so can [pass in](https://github.com/App-vNext/Polly/blob/2862be4f4db5fd2fe4a97e9c55faf4b49793e95d/src/Polly.Shared/PolicyAsync.cs#L469)  a `Context`.  In the `HttpClientFactory` integration, that call is (necessarily) [within `PolicyHttpMessageHandler.cs`](https://github.com/aspnet/HttpClientFactory/blob/ad8554ace8957a371523dad8395a220d4aec3f26/src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs#L33), so the user cannot directly pass in `Context`.

`HttpClient` being (presumably) not for modification, [`HttpRequestMessage.Properties`](https://github.com/dotnet/corefx/blob/bffef76f6af208e2042a2f27bc081ee908bb390b/src/System.Net.Http/src/System/Net/Http/HttpRequestMessage.cs#L113-L123) seems like the intended(?)/available place for attaching execution-scoped data.  

The extra line, [5], would not be necessary for ordinary usage.  It would be an available option, for those who wanted the richer telemetry or to use `CachePolicy`.   

---

Thoughts? Should this be part of the Polly-HttpClientFactory story? Are there any good alternatives?

rynowak:
Resolves #62


src/Microsoft.Extensions.Http.Polly/HttpRequestMessageExtensions.cs
```diff
@@ -0,0 +1,58 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Http;
+using System;
+using System.Net.Http;
+
+namespace Polly
```
rynowak: My goal is to avoid polluting System.Net.Http. We could also use `Microsoft.Extensions.Http` but it seems more natural to use the Polly namespace since you'd likely have it already added if you're interested in the context. Thoughts?

src/Microsoft.Extensions.Http.Polly/HttpRequestMessageExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// Gets the <see cref="Context"/> associated with the provided <see cref="HttpRequestMessage"/>.
+        /// </summary>
+        /// <param name="request">The <see cref="HttpRequestMessage"/>.</param>
+        /// <returns>The <see cref="Context"/> is set, otherwise <c>null</c>.</returns>
+        /// <remarks>
+        /// The <see cref="PolicyHttpMessageHandler"/> will attach a context to the <see cref="HttpResponseMessage"/> prior
+        /// to executing a <see cref="Policy"/>, if one does not already exist. The <see cref="Context"/> will be provided
+        /// to the policy for use inside the <see cref="Policy"/> and in other message handlers.
+        /// </remarks>
+        public static Context GetPollyContext(this HttpRequestMessage request)
```
rynowak: Thoughts on naming? I also considered `GetPolicyContext`, which I kinda like better on second thought.

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
```diff
@@ -77,7 +82,38 @@ protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage reques
                 throw new ArgumentNullException(nameof(request));
             }
 
-            return _policy.ExecuteAsync((ct) => base.SendAsync(request, ct), cancellationToken);
+            // Guarantee the existance of a context for every policy execution, but only create a new one if needed. This
+            // allows later handlers to flow state if desired.
+            var context = request.GetPollyContext();
+            if (context == null)
+            {
+                context = new Context(Guid.NewGuid().ToString());
```
rynowak: Suggestions how to create the executioncontextid?

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
```diff
@@ -77,7 +82,38 @@ protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage reques
...
+                request.SetPollyContext(context);
+            }
+
+            return _policy.ExecuteAsync((c, ct) => SendCoreAsync(request, c, ct), context, cancellationToken);
```
rynowak: Note that I'm explicitly **not** cleaning up the context here. It's not clear that there's any value to that, but I'm interested in thoughts.

I have a suspicion that we might want to... but it would only affect some truly weird scenarios. Like if someone has implemented their own retry support, but they aren't setting the context themself.

src/Microsoft.Extensions.Http.Polly/HttpRequestMessageExtensions.cs
rynowak: We could also give this a scarier name like `GetPolicyExecutionContext` to act as salt.

src/Microsoft.Extensions.Http.Polly/HttpRequestMessageExtensions.cs
rynowak: With the metapackage approach, it means that a lot of users have will have this package available without realizing it. This is ultimately going to be included in the `Microsoft.AspNetCore.All` package.

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
Yves57: @rynowak Just for curiosity, why is there always a `Context` creation for each request execution (with `Context` and `string` allocation), even for uses that don't use it? Why didn't you prefer something like
```
public static Context GetPollyContext(this HttpRequestMessage request)
{
   if (!request.Properties.TryGetValue(PollyContextKey, out var context))
   {
       context = new Context(Guid.NewGuid().ToString());
       request.Properties[PollyContextKey] = context;
   }
   return context;
}
```

martincostello: `GetPollyContext()` is doing the `TryGet()` here and `SetPollyContext()` is going the set when one needs to be made. In that example it's not really getting, because it sets if it's not already there. Your example would be more `EnsurePollyContext()`.

src/Microsoft.Extensions.Http.Polly/HttpRequestMessageExtensions.cs
reisenberger: Re:
>goal is to avoid polluting System.Net.Http. [...] seems more natural to use the Polly namespace

:+1: So the extension methods would not appear unless someone had `using Polly;`, not confusing those not interested.

src/Microsoft.Extensions.Http.Polly/HttpRequestMessageExtensions.cs
reisenberger: My vote on naming is for `GetPolicyExecutionContext()`: reinforces the idea that the context is execution-scoped.  If so, should the change flow elsewhere also?, eg:

    internal static readonly string PolicyExecutionContextKey = "PolicyExecutionContext";

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
reisenberger: >Suggestions how to create the executioncontextid?

If the user has attached no context to the request, just:

    context = new Context();

[`Context.ExecutionKey`](https://github.com/App-vNext/Polly/blob/754a4a920900e1ae56d4fdae7cbdfd02bcb8cbd1/src/Polly.Shared/Context.cs#L41-L45) is the key [set via constructor](https://github.com/App-vNext/Polly/blob/754a4a920900e1ae56d4fdae7cbdfd02bcb8cbd1/src/Polly.Shared/Context.cs#L21-L24).  This represents the call site usage of the policy.  Not-set is allowed; if the user has not chosen to tag Polly logging/telemetry by call site tags (by using `request.SetPollyContext(new Context("MyUsageTag"));`),  we should not set it to something arbitrary.

_Compare:_ [`Context.ExecutionGuid`](https://github.com/App-vNext/Polly/blob/754a4a920900e1ae56d4fdae7cbdfd02bcb8cbd1/src/Polly.Shared/Context.cs) is the separate, per-execution-unique correlation id. That is [automatically initialised by Polly](https://github.com/App-vNext/Polly/blob/754a4a920900e1ae56d4fdae7cbdfd02bcb8cbd1/src/Polly.Shared/Context.cs#L51-L58).

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
reisenberger: >explicitly not cleaning up the context here

It seems correct (in fact important) that we don't clean up the context.  This line in execution could be hit within the looping flowing from an enclosing retry policy.  We'd want the same context to hold through all retries, so we shouldn't (for example) set it to null here.  (That iiuc your thought re 'clean up context' - come back on this if I'm missing something, or if you see scenarios that might contradict that!)

test/Microsoft.Extensions.Http.Polly.Test/PolicyHttpMessageHandlerTest.cs
```diff
@@ -103,8 +97,65 @@ public class PolicyHttpMessageHandlerTest
...
+            // Act
+            var response = await handler.SendAsync(request, CancellationToken.None);
+
+            // Assert
+            Assert.Same(context, request.GetPollyContext()); // We don't clean up the context
+            Assert.Same(expected, response);
+        }
+
+        [Fact]
+        public async Task SendAsync_ExistingContext_ReusesContext()
```
reisenberger: Test as is does not check that the correct context is passed to SendAsync?  SUT could push any other context onto `HttpRequestMessage` before calling `SendAsync(...)`, pop it off again after, and the test would still pass?  Could strengthen by tracking `Context contextPassedToSendAsync;`

test/Microsoft.Extensions.Http.Polly.Test/PolicyHttpMessageHandlerTest.cs
```diff
@@ -103,8 +97,65 @@ public class PolicyHttpMessageHandlerTest
...
+        {
+            // Arrange
+            var policy = Policy.TimeoutAsync<HttpResponseMessage>(TimeSpan.FromSeconds(10));
+            var handler = new TestPolicyHttpMessageHandler(policy);
+
+            var expected = new HttpResponseMessage();
+            handler.OnSendAsync = (req, c, ct) =>
+            {
+                Assert.NotNull(c);
+                Assert.Same(c, req.GetPollyContext());
```
reisenberger: `contextPassedToSendAsync = c;`

test/Microsoft.Extensions.Http.Polly.Test/PolicyHttpMessageHandlerTest.cs
```diff
@@ -103,8 +97,65 @@ public class PolicyHttpMessageHandlerTest
...
+            };
+
+            var request = new HttpRequestMessage();
+            var context = new Context(Guid.NewGuid().ToString());
+            request.SetPollyContext(context);
+
+            // Act
+            var response = await handler.SendAsync(request, CancellationToken.None);
+
+            // Assert
```
reisenberger: `Assert.AreSame(context, contextPassedToSendAsync);`

src/Microsoft.Extensions.Http.Polly/HttpRequestMessageExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+using System;
+using System.Net.Http;
+
+namespace Polly
+{
+    /// <summary>
+    /// Extension methods for <see cref="HttpRequestMessage"/> Polly integration.
+    /// </summary>
+    public static class HttpRequestMessageExtensions
+    {
```
reisenberger: Awesome to see the `Context` support going in; great that we can add xtn methods to `HttpRequestMessage` to hide the details.

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
reisenberger: @Yves57 Re:

>why is there always a Context creation for each request execution [...] even for uses that don't use it? 
>Why didn't you prefer [...] so the context is allocated only if the user ask for it?

Polly uses a `Context` any time executing through a policy, to provide context to telemetry and logging, so the allocation isn't extra; Polly would allocate it later internally if it wasn't done here.

By scoping the context to the `HttpRequestMessage`, we ensure that the same `Context` instance is used through the whole chain of `DelegatingHandler`s when the user has configured multiple `PolicyHttpMessageHandler` instances.  This might be common, for example using both a circuit-breaker and retry policy: `.AddPolicyHandler(/* some circuit-breaker */).AddPolicyHandler(/* some retry policy */)`.  

While the motivation was to make that `Context` consistent through a chain of `PolicyHttpMessageHandler`s, it will actually mean fewer allocations: if `Context` wasn't scoped to `HttpRequestMessage`, the execution through each policy would generate a fresh `Context`, meaning more allocations.

Good question; hope this explanation helps.

Yves57: @reisenberger Thank you for the precise answer! Looking in the Polly code, I have seen that the "default" context that would be allocated. So I understand now that the best is to allocate it here.

src/Microsoft.Extensions.Http.Polly/HttpRequestMessageExtensions.cs
rynowak: Sure, will make this change.

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
rynowak: @reisenberger - FYI it will be `new Context(null)` the no-arg version is internal.

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
rynowak: Ok good, sounds like this is right then.

test/Microsoft.Extensions.Http.Polly.Test/PolicyHttpMessageHandlerTest.cs
rynowak: Sure, this got missed somehow.

rynowak:
Updated and addressed feedback. If there's nothing else important to take action on here, I plan to merge this tomorrow.

test/Microsoft.Extensions.Http.Polly.Test/PolicyHttpMessageHandlerTest.cs
rynowak: This comment is a bit stale now, but I addressed the underlying problem

test/Microsoft.Extensions.Http.Polly.Test/PolicyHttpMessageHandlerTest.cs
rynowak: This comment is a bit stale now, but I addressed the underlying problem

PR #66, opened 2018-02-24
Add support for policy registry

rynowak:
Here's a straw man to discuss for the design of retrieving policies from
the registry for IHttpClientFactory.

Goals:
 - Keep simple usage simple
 - Support updates (opt in)
 - Allow use of policies by name and with code
 - Allow dynamism (choosing policy based on request)

 Non-goal:
 - Avoid overloading based on Func<> types with same arity. This is a
 destructive case for intellisense and should be avoided.


samples/HttpClientFactorySample/Program.cs
```diff
@@ -58,6 +66,24 @@ public static void Configure(IServiceCollection services)
...
+                return request.Method == HttpMethod.Get ? timeout : longTimeout;
+            })
+
+            // Run some code to select a policy from the registry based on the request
+            .AddPolicyHandlerFromRegistry((reg, request) =>
+            {
+                return request.Method == HttpMethod.Get ?
+                    reg.Get<IAsyncPolicy<HttpResponseMessage>>("regular") :
+                    reg.Get<IAsyncPolicy<HttpResponseMessage>>("long");
+            })
```
rynowak: This is the best place to start looking. Let's try to nail down what the configuration experience is like and then let the actual implementation take care of itself 😆

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
```diff
@@ -74,6 +75,96 @@ public static IHttpClientBuilder AddPolicyHandler(this IHttpClientBuilder builde
...
+            {
+                var registry = services.GetRequiredService<IReadOnlyPolicyRegistry<string>>();
+
+                // The policy might be either IAsyncPolicy or IAsyncPolicy<HttpRequestMessage> but never both.
+                // But, the handler expects IAsyncPolicy<HttpRequestMessage>, so try first for what we want and then
+                // fall back to the non-generic interface. We'll get the non-generic case throw the policy isn't registred.
+                if (!registry.TryGet<IAsyncPolicy<HttpResponseMessage>>(policyKey, out var policy))
+                {
+                    policy = registry.Get<IAsyncPolicy>(policyKey).WrapAsync(Policy.NoOpAsync<HttpResponseMessage>());
+                }
```
rynowak: @reisenberger - there's a lot of tension caused in this design by the desire to support `IAsyncPolicy` and `IAsyncPolicy<HttpResponseMessage>`.

For instance if we were to support `Func<HttpRequestMessage, IAsyncPolicy<HttpResponseMessage>>` **and** `Func<HttpRequestMessage, IAsyncPolicy>`, then we end up with a method that's overload on the return type of `Func<..T>`. In general this case gives really bad intellisense and I try to avoid it.

It seems like it's always possible to create a `PolicyAsync<T>` instead of a `PolicyAsync` for anything you desire. Thoughts?

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+using System;
+using Polly.Registry;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    /// <summary>
+   /// Provides convenience extension methods to register <see cref="IPolicyRegistry{String}"/> and 
+   /// <see cref="IReadOnlyPolicyRegistry{String}"/> in the service collection.
+    /// </summary>
+    public static class PollyServiceCollectionExtensions
```
rynowak: These are just for convenience. Our `AddPolicyHandlerFromRegistry` methods will require you to have registered `IReadOnlyPolicyRegistry<string>` somehow. We don't care if you used these methods to do it.

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        /// <summary>
+        /// Registers and empty <see cref="PolicyRegistry"/> in the service collection with service types
```
martincostello: Typo: _"Registers **an** empty."_

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
```diff
@@ -77,7 +92,13 @@ protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage reques
                 throw new ArgumentNullException(nameof(request));
             }
 
-            return _policy.ExecuteAsync((ct) => base.SendAsync(request, ct), cancellationToken);
+            var policy = ResolvePolicy(request);
+            return policy.ExecuteAsync((ct) => base.SendAsync(request, ct), cancellationToken);
```
martincostello: Should there be a null-check on `policy` after `ResolvePolicy()` returns?

martincostello: Does there need to be a mechanic here to provide a way to create a `Context` which could be passed into the policy, or is that going to be catered for separately/elsewhere? Given that the policy is already being resolved from the request, there could be a hook to resolve a context from the request as well.

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
```diff
@@ -77,7 +92,13 @@ protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage reques
...
+        }
+
+        private IAsyncPolicy<HttpResponseMessage> ResolvePolicy(HttpRequestMessage request)
```
bryanjhogan: @rynowak this is great, I think it solves most of the things I was concerned about.

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
reisenberger: >a lot of tension caused in this design by the desire to support `IAsyncPolicy` and `IAsyncPolicy<HttpResponseMessage>`.

Supporting only `IAsyncPolicy<HttpResponseMessage>` would certainly simplify the implementation.  With `HttpClientFactory` there a better case for supporting only generic policies than with Polly in general, given all calls are in `HttpResponseMessage`.

A concern is that not supporting non-generic `IAsyncPolicy` will lead to enquiries of why, eg, `.AddPolicyHandler(Policy.TimeoutAsync(10))` doesn't compile. In Polly's API, I opted to take the hit and provide bridging overloads ([eg in PolicyWrap](https://github.com/App-vNext/Polly/blob/754a4a920900e1ae56d4fdae7cbdfd02bcb8cbd1/src/Polly.Shared/Wrap/PolicyWrapSyntaxAsync.cs#L44-L51)), so that from the user's perspective, "everything just worked".  'Traffic-managing' resilience strategies like Timeout and Bulkhead have no interest in returned results, so we opted to ensure that non-generic forms could [just be used](https://github.com/App-vNext/Polly/wiki/Non-generic-and-generic-policies#mixing-non-generic-and-generic-in-a-policywraptresult) wherever a generic form might otherwise be expected.

If the decision is that `HttpClientFactory` only supports `IAsyncPolicy<HttpResponseMessage>`, documentation should cover this, to reduce user blocks, and enquiries.

---

Polly could add a helper extension method `Policy<TResult> Policy.For<TResult>()` (the equivalent of your `.WrapAsync(Policy.NoOpAsync<HttpResponseMessage>())` without the need for the extra `NoOpPolicy<TResult>`).  This could either help users bridge; or `HttpClientFactory` provide the extra overloads.

---

Worth further thought and discussion, as will be key to the user experience.

rynowak: I had a discussion this morning about this with @glennc - we're thinking that we want to try removing/not-providing the non-generic methods for the preview release and see if we get feedback.

There's a relatively small delta between `.DoThing(Policy.Timeout(...))` and `.DoThing(Policy.Timeout<HttpResponseMessage>())`. We'd rather require the extra typing than ship partial or complicated support for non-generic policies.

If we get the feedback that this is important then we will do the work to add overloads that directly accept non-generic `IAsyncPolicy`.

--

I agree that it would be a nice addition to have a simpler way to go from `IAsyncPolicy`->`IAsyncPolicy<>`. I suppose that's on you guys if you think it's worth doing. For now I plan to just document how to do it.

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
rynowak: I agonized over this and couldn't really come to a good decision. Usually my philosophy is *play stupid games, win stupid prizes*... 

This can be done in a way that doesn't impact the perf of the non-dynamic way so it's probably worth doing to allay confusion.

rynowak: See https://github.com/aspnet/HttpClientFactory/pull/65

Whomever is the caller can provide a context, or another message handler can do it (since they are also callers).

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
```diff
@@ -74,6 +75,96 @@ public static IHttpClientBuilder AddPolicyHandler(this IHttpClientBuilder builde
             return builder;
         }
 
+        public static IHttpClientBuilder AddPolicyHandler(
+            this IHttpClientBuilder builder, 
+            Func<HttpRequestMessage, IAsyncPolicy<HttpResponseMessage>> policyFactory)
```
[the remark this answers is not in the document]
rynowak: Sure, that sounds fine.

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
```diff
@@ -109,7 +200,16 @@ public static IHttpClientBuilder AddPolicyHandler(this IHttpClientBuilder builde
             this IHttpClientBuilder builder,
```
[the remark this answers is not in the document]
rynowak: I think this feedback makes sense. I'm going to consider this out of scope for this PR thought, will address it in a follow up.

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
bryanjhogan: I think this is a balanced approach

We don't really know whether people will miss the non-generic methods so it's probably fine to leave it out and see who complains.

src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
```diff
@@ -77,7 +92,13 @@ protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage reques
...
+        {
+            return _policy ?? _policyFactory(request);
```
bryanjhogan: @rynowak How do you plan to make to `PolicyRegistry` accessible from this class?

rynowak: I don't plan to do that. You provide a lambda that resolves the policy, take a look at the sample and let me know if that won't meet your needs.

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
reisenberger: @rynowak Very happy with that call :+1:; I agonised over this. 

Offering the mix of non-generic and generic overloads makes sense for Polly generally, as we have a range of use cases to support, some of which suit non-generic policies better; but with `HttpClientFactory`'s one use case (for Polly), there is a chance to reduce complexity.  Simplicity is a great virtue.

rynowak:
I've addressed all of the minor feedback here. Planning to add tests tomorrow

test/Microsoft.Extensions.Http.Test/DefaultHttpClientFactoryTest.cs
```diff
@@ -246,7 +246,7 @@ public void Factory_CreateClient_FiltersCanDecorateBuilder()
 
             // Assert - 2
             Assert.Empty(factory._activeHandlers);
-            Assert.True(factory.CleanupTimerStarted.IsSet);
+            Assert.True(factory.CleanupTimerStarted.IsSet, "Cleanup timer started");
```
rynowak: This is just misc cleanup that snuck in while I was looking at a flaky test 😆

rynowak:
Updated and added tests. I converted the existing builder tests into end-to-ends.

Issue #69, opened 2018-03-02
Bug in HttpClientFactorySample RetryHandler causes no time between retries


skleanthous:
### Title
In the `RetryHandler` class, in the `SendAsync` method, when an exception occurs, a call to `Task.Delay` is made but the result is never used/awaited.

### Functional impact
The sample code of RetryHandler will not wait for retries on error, instead it will instantly retry.

### Further technical details
See code here: https://github.com/aspnet/HttpClientFactory/blob/dev/samples/HttpClientFactorySample/Program.cs#L108

skleanthous:
More than eager to create a PR to fix this :)

Tratcher:
Go right ahead. We'll get the CLA sorted out.

rynowak:
Thanks!

PR #70, opened 2018-03-02
Await Task.Delay in RetryHandler in HttpClientFFactorySample

skleanthous:
- Changed `SendAsync` method of `RetryHandler` in `HttpClientFactorySample` to be async.
- await instead of return SendAsync
- await result of Task.Delay

Addresses #69


rynowak:
Thanks!

PR #78, opened 2018-03-21
Design strawman for transient errors

rynowak:
A design strawman showing what an opinionated 'transient errors' set of
extensions would look like.

### DIFF
--- a/src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
+++ b/src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
@@ -44,45 +44,87 @@ namespace Microsoft.Extensions.DependencyInjection
             return builder;
         }
 
+        /// <summary>
+        /// Adds a <see cref="PolicyHttpMessageHandler"/> which will surround request execution with a policy returned
+        /// by the <paramref name="policySelector"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="IHttpClientBuilder"/>.</param>
+        /// <param name="policySelector">
+        /// Selects an <see cref="IAsyncPolicy{HttpResponseMessage}"/> to apply to the current request.
+        /// </param>
+        /// <returns>An <see cref="IHttpClientBuilder"/> that can be used to configure the client.</returns>
+        /// <remarks>
+        /// <para>
+        /// See the remarks on <see cref="PolicyHttpMessageHandler"/> for guidance on configuring policies.
+        /// </para>
+        /// </remarks>
         public static IHttpClientBuilder AddPolicyHandler(
             this IHttpClientBuilder builder, 
-            Func<HttpRequestMessage, IAsyncPolicy<HttpResponseMessage>> policyFactory)
+            Func<HttpRequestMessage, IAsyncPolicy<HttpResponseMessage>> policySelector)
         {
             if (builder == null)
             {
                 throw new ArgumentNullException(nameof(builder));
             }
 
-            if (policyFactory == null)
+            if (policySelector == null)
             {
-                throw new ArgumentNullException(nameof(policyFactory));
+                throw new ArgumentNullException(nameof(policySelector));
             }
 
-            builder.AddHttpMessageHandler(() => new PolicyHttpMessageHandler(policyFactory));
+            builder.AddHttpMessageHandler(() => new PolicyHttpMessageHandler(policySelector));
             return builder;
         }
 
+        /// <summary>
+        /// Adds a <see cref="PolicyHttpMessageHandler"/> which will surround request execution with a policy returned
+        /// by the <paramref name="policySelector"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="IHttpClientBuilder"/>.</param>
+        /// <param name="policySelector">
+        /// Selects an <see cref="IAsyncPolicy{HttpResponseMessage}"/> to apply to the current request.
+        /// </param>
+        /// <returns>An <see cref="IHttpClientBuilder"/> that can be used to configure the client.</returns>
+        /// <remarks>
+        /// <para>
+        /// See the remarks on <see cref="PolicyHttpMessageHandler"/> for guidance on configuring policies.
+        /// </para>
+        /// </remarks>
         public static IHttpClientBuilder AddPolicyHandler(
             this IHttpClientBuilder builder,
-            Func<IServiceProvider, HttpRequestMessage, IAsyncPolicy<HttpResponseMessage>> policyFactory)
+            Func<IServiceProvider, HttpRequestMessage, IAsyncPolicy<HttpResponseMessage>> policySelector)
         {
             if (builder == null)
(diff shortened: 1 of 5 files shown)


src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
```diff
@@ -44,45 +44,87 @@ public static IHttpClientBuilder AddPolicyHandler(this IHttpClientBuilder builde
...
+        /// <returns>An <see cref="IHttpClientBuilder"/> that can be used to configure the client.</returns>
+        /// <remarks>
+        /// <para>
+        /// See the remarks on <see cref="PolicyHttpMessageHandler"/> for guidance on configuring policies.
+        /// </para>
+        /// </remarks>
         public static IHttpClientBuilder AddPolicyHandler(
             this IHttpClientBuilder builder, 
-            Func<HttpRequestMessage, IAsyncPolicy<HttpResponseMessage>> policyFactory)
+            Func<HttpRequestMessage, IAsyncPolicy<HttpResponseMessage>> policySelector)
```
rynowak: Missed this rename earlier and added docs where they were missing

src/Microsoft.Extensions.Http.Polly/PolicyExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+        public static PolicyBuilder<HttpResponseMessage> OrTransientHttpStatusCode(this PolicyBuilder policyBuilder)
+        {
+            if (policyBuilder == null)
+            {
+                throw new ArgumentNullException(nameof(policyBuilder));
+            }
+
+            return policyBuilder.OrResult<HttpResponseMessage>(response =>
+            {
+                return (int)response.StatusCode >= 500 || response.StatusCode == HttpStatusCode.RequestTimeout;
```
martincostello: Could `429` be considered "transient", or is that something that would be handled by something more specialised?

rynowak: Good question, I think for now we're intentionally leaving it out. You'd want to handle that differently than a timeout or a transient connection issue.

reisenberger: Agree about probably not including `429` here:

+ We can expect users to use `OrTransientHttpStatusCode()` with policies other than Retry, for example CircuitBreaker.  We may not want to embed _as a default_ that `429` is bad enough news to contribute towards breaking a circuit. 

+ As you say @martincostello and @rynowak , 429 can play nicely with specialised handling.  For example Polly has [WaitAndRetry(...) overloads](https://github.com/App-vNext/Polly/wiki/Retry#retryafter-when-the-response-specifies-how-long-to-wait) enabling the policy to wait before retrying for exactly the duration specified in any `Retry-After` or `x-ms-retry-after-ms` header.  If we embed `429` in `OrTransientHttpStatusCode()`, we force users to handle 429 and other retry-able causes in the same policy, possibly leading to internal forking logic and limiting options for treating them differently.   

`429` could be covered by some documentation examples.

src/Microsoft.Extensions.Http.Polly/PolicyExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+        public static PolicyBuilder<HttpResponseMessage> OrTransientHttpError(this PolicyBuilder policyBuilder)
+        {
+            if (policyBuilder == null)
+            {
+                throw new ArgumentNullException(nameof(policyBuilder));
+            }
+
+            return policyBuilder.Or<HttpRequestException>().OrTransientHttpStatusCode();
+        }
+    }
```
reisenberger: From where I'm sitting the exceptions and status codes covered by the opinionated methods here feels like a coherent set :+1: (tho intd in other views).  Users can also extend the reach of exceptions or status codes the policy handles with Polly's fluent syntax.

---

We might want to add the following additional two overloads:

```csharp
    public static PolicyBuilder<HttpResponseMessage> OrTransientHttpStatusCode(this PolicyBuilder<HttpResponseMessage> policyBuilder)
        {
            if (policyBuilder == null)
            {
                throw new ArgumentNullException(nameof(policyBuilder));
            }

            return policyBuilder.OrResult<HttpResponseMessage>(response =>
            {
                return (int)response.StatusCode >= 500 || response.StatusCode == HttpStatusCode.RequestTimeout;
            });
        }

        public static PolicyBuilder<HttpResponseMessage> OrTransientHttpError(this PolicyBuilder<HttpResponseMessage> policyBuilder)
        {
            if (policyBuilder == null)
            {
                throw new ArgumentNullException(nameof(policyBuilder));
            }

            return policyBuilder.Or<HttpRequestException>().OrTransientHttpStatusCode();
        }
```

_Explanation:_ If the user starts defining a policy with:

    Policy.HandleResult<HttpResponseMessage>(r => r.StatusCode == /* some other status code they also want to handle */)

then they have a `PolicyBuilder<HttpResponseMessage>` right off the bat.  So, for completeness, the above provides variants of `OrTransientHttpStatusCode(...)` and `OrTransientHttpError(...)` which also chain fluently on `PolicyBuilder<HttpResponseMessage>`.

(guess it would then also be an option to factor out some common code across the overloads)

reisenberger:
PR LGTM.  No other comment except spotted the odd intellisense typo - do you want a review on this PR listing those, or just a separate PR to fix?

rynowak:
@reisenberger - if you have any minor feedback about this PR bring it. I think all of the places I added documentation and going to be finalized

reisenberger (commented):
Handful of small typos

src/Microsoft.Extensions.Http.Polly/PolicyExtensions.cs
rynowak: Makes sense.

rynowak:
Updated

src/Microsoft.Extensions.Http.Polly/DependencyInjection/PollyHttpClientBuilderExtensions.cs
```diff
@@ -148,14 +204,12 @@ public static IHttpClientBuilder AddPolicyHandlerFromRegistry(this IHttpClientBu
...
-        /// {
-        ///     return response.StatusCode >= HttpStatusCode.InternalServerError;
-        /// }
-        /// </code>
+        /// preconfigured errors in the following categories:
+        /// <list type="bullet">
+        /// <item><description>Network failures (as <see cref="HttpRequestException"/>)</description></item>
+        /// <item><description>HTTP 5XX status codes (server errors)</description></item>
+        /// <item><description>HTTP 408 status code (request timeout)</description></item>
+        /// </list>
```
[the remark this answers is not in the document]
rynowak: I think I fixed these.

rynowak:
I'm going to go ahead and merge this for the 2.1 preview2 release to minimize churn. If there's a Polly.Extensions.Http package available before we have to lock down I'll do the work to integrate it.

PR #79, opened 2018-03-22
Remove support for NonGeneric Polly policies from PolicyRegistry helper methods

reisenberger:
Removes support for NonGeneric Polly policies from `IHttpClientBuilder.AddPolicyHandlerFromRegistry(...)`

We decided that `HttpClientFactory` would not support Polly's non-generic policies `IAsyncPolicy` at this time.  This PR removes one left-over reference.


rynowak:
Ooh, I didn't realise I had missed this one. Thanks! Will merge when CI is green

rynowak:
This change doesn't build, looks like it needs a `var`. I'll get around to it in the next day or so if you don't have a chance to fix it.

rynowak:
The test failure on mac is a known VS test issue, so I'm ignore it. Thanks for the help cleaning this up!

PR #82, opened 2018-03-23
Fix #75 Add HttpClient+IServiceProvider methods

### LINKED ISSUE #75: Add overloads with IServiceProvider to extension methods?
As part of preliminary investigation of how to update the documentation and samples for `JustEat.HttpClientInterception` for ASP.NET Core 2.1 and HttpClientFactory (see justeat/httpclient-interception#23), I found that there was a lack of overloads on some extension method types that provided a ```IServiceProvider``` for some scenarios.

While not the end of the world in this particular example, I think this might be more of an issue in an application where settings from strongly-typed configuration (i.e. `Options<T>` things) are wanted by the consuming app to augment the HTTP client builder.

Is there merit in adding overloads similar to the following to allow for such usage patterns?

```csharp
// HttpClientBuilderExtensions
public static IHttpClientBuilder ConfigureHttpClient(this IHttpClientBuilder builder, Action<HttpClient, IServiceProvider> configureClient);

// HttpClientFactoryServiceCollectionExtensions
public static IHttpClientBuilder AddHttpClient(this IServiceCollection services, string name, Action<HttpClient, IServiceProvider> configureClient);
```

This question might be more of a question of discoverability, rather than missing functionality, and I'm just looking at things wrong. For example, I later found an overload to `AddTypedClient()` I missed initially, which lead to another refactor ([commit](https://github.com/justeat/httpclient-interception/pull/23/commits/a76c5edd33e20f315ef2d22aa313266ae88a969e)).

rynowak:
Adds extension methods that allow you to accept an IServiceProvider when
configuring HttpClient.

### DIFF
--- a/src/Microsoft.Extensions.Http/DependencyInjection/HttpClientBuilderExtensions.cs
+++ b/src/Microsoft.Extensions.Http/DependencyInjection/HttpClientBuilderExtensions.cs
@@ -38,6 +38,35 @@ namespace Microsoft.Extensions.DependencyInjection
         }
 
         /// <summary>
+        /// Adds a delegate that will be used to configure a named <see cref="HttpClient"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="IServiceCollection"/>.</param>
+        /// <param name="configureClient">A delegate that is used to configure an <see cref="HttpClient"/>.</param>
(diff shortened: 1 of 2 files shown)


rynowak:
/cc @martincostello let me know if this matches your expectations 👍 

One limitation here is that this will be an **application level** service provider. You won't get access to scoped or per-request services.

PR #91, opened 2018-04-06
Adding HttpClient to the log category name

### LINKED ISSUE #90: Include HttpClient name in log messages
This is a follow up from #51.

In that issue @andrewlock suggest log messages be prefixed `System.Net.Http.HttpClient.` rather than `System.Net.Http.`.

@glennc agreed and @rynowak seems to suggest this was implemented but I don't see that in code?

Happy to submit a PR for the change if it is?

stevejgordon:
Fixes #90


stevejgordon:
No problem! Caught it while checking stuff for the docs. :-)

Issue #94, opened 2018-04-11
Option: Selectively clean up Polly Context?


reisenberger:
Revisiting [this conversation](https://github.com/aspnet/HttpClientFactory/pull/65#discussion_r170357438) (click also "Show outdated") about cleaning up Polly execution `Context`.  I don't think we necessarily have to change anything - the below doesn't affect any recommended HttpClientFactory usage pattern afaics - but I'm airing an option so we've had the discussion.

---

Where we currently [do this](https://github.com/aspnet/HttpClientFactory/blob/60a23380a354ebaf27c8cd3505e745bd8b7f1541/src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs#L117-L125), we could:

```csharp
var cleanUpContext = false;
var context = request.GetPolicyExecutionContext();
if (context == null)
{
    context = new Context();
    request.SetPolicyExecutionContext(context);
    cleanUpContext = true;
}

var policy = _policy ?? SelectPolicy(request);
HttpResponseMessage response = await policy.ExecuteAsync((c, ct) => SendCoreAsync(request, c, ct), context, cancellationToken);

if (cleanUpContext)
{
    request.SetPolicyExecutionContext(null);
}

return response;
```

We cannot clean up `Context` on exiting every `PolicyHttpMessageHandler`. We need to flow `Context` inwards and outwards through nested `PolicyHttpMessageHandler`s so that telemetry for an entire execution (including retries) correlates with the unique [`CorrelationId` on `Context`](https://github.com/App-vNext/Polly/wiki/Keys-and-Context-Data#correlationid).

But if we created the `Context`, we know both that it was not user-supplied, and that we are the outermost `PolicyHttpMessageHandler`, so we could clean it up.

**Philosophically** this feels slightly cleaner, in adhering to the principle of least side-effect: whether the overall call thru `HttpClient` was supplied an `HttpRequestMessage` with or without context, we return it in the same state. It's possible this extra cleanliness could prevent us being bitten later in some as-yet-unforeseen case.

[...]

PR #96, opened 2018-04-13
Selectively clean up Polly Context

reisenberger:
Clean up the Polly execution `Context` on exit from the outermost `PolicyHttpMessageHandler`, if that `Context` was not user-supplied.

Addresses #94


src/Microsoft.Extensions.Http.Polly/PolicyHttpMessageHandler.cs
```diff
@@ -114,15 +114,24 @@ protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage reques
...
             var policy = _policy ?? SelectPolicy(request);
-            return policy.ExecuteAsync((c, ct) => SendCoreAsync(request, c, ct), context, cancellationToken);
+            var response = await policy.ExecuteAsync((c, ct) => SendCoreAsync(request, c, ct), context, cancellationToken);
+
+            if (cleanUpContext)
+            {
+                request.SetPolicyExecutionContext(null);
+            }
+
+            return response;
```
rynowak: Should this be in a try/finally?

reisenberger: :+1: This isn't about releasing critical resources, but it does make the clean-up guarantee stronger.

Added this to the PR, with tests.  Also covered the case where _policySelector throws.

PR #98, opened 2018-04-13
Redesign ITypedClientFactory to be generic

rynowak:
This gives us a much better caching behavior for using
ActivatorUtilities and will avoid reflection on the common code path.

### DIFF
--- a/src/Microsoft.Extensions.Http/DefaultTypedHttpClientFactory.cs
+++ b/src/Microsoft.Extensions.Http/DefaultTypedHttpClientFactory.cs
@@ -3,13 +3,19 @@
 
 using System;
 using System.Net.Http;
+using System.Threading;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.Extensions.Http
 {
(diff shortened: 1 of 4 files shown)


src/Microsoft.Extensions.Http/DefaultTypedHttpClientFactory.cs
```diff
@@ -3,14 +3,20 @@
...
 using System.Net.Http;
+using System.Threading;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.Extensions.Http
 {
-    internal class DefaultTypedHttpClientFactory : ITypedHttpClientFactory
+    internal class DefaultTypedHttpClientFactory<TClient> : ITypedHttpClientFactory<TClient>
     {
+        private readonly static Func<ObjectFactory> _createActivator = () => ActivatorUtilities.CreateFactory(typeof(TClient), new Type[] { typeof(HttpClient), });
```
pakrym: What if this throws? Where would exception surface?

rynowak:
I saw you delete that comment you sneaky pete.

PR #100, opened 2018-04-16
Improve formatting and key-value logs

rynowak:
This change improves the logging of request and response headers for
both the command line and integration with structured logging systems
like seq.

Each request (at trace level) will have all headers logged as key-value
pairs.


rynowak:
@davidfowl you were right

src/Microsoft.Extensions.Http/Logging/HttpHeadersLogValue.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return Values.GetEnumerator();
+        }
+
+        public override string ToString()
+        {
+            return _kind.ToString();
+        }
```
rynowak: Overriding `ToString()` to do something less useful. SEQ likes to ToString() things and record that in addition to the key-value pairs.

src/Microsoft.Extensions.Http/Logging/HttpHeadersLogValue.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        public HttpHeaders ContentHeaders { get; }
+        
+        private List<KeyValuePair<string, object>> Values
+        {
+            get
+            {
+                if (_values == null)
+                {
+                    var values = new List<KeyValuePair<string, object>>();
```
rynowak: We don't really have a choice but to allocate. The `HttpHeaders` types are pretty opaque.

This only happens at Trace anyway

src/Microsoft.Extensions.Http/Logging/HttpHeadersLogValue.cs
davidfowl: I think we moved to using ToString for most things instead of a Format method (https://github.com/aspnet/Hosting/blob/a997ceba98c40a98d2fd48190cba6d12abf105e7/src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs#L137). 

Why not put this logic in ToString instead?

davidfowl:
Do you have a sample of what this looks like in seq?

src/Microsoft.Extensions.Http/Logging/HttpHeadersLogValue.cs
rynowak: Yeah it turns out this doesn't really make a difference

rynowak:
image

davidfowl:
That's pretty darn beautiful.

PR #102, opened 2018-04-25
Private field '_expiryCallback' is assigned but never used

tangdf:


martincostello:
Could it be that it's supposed to be being used here: https://github.com/tangdf/HttpClientFactory/blob/0c3c4b3aa06b451ab52a0e5610cca3d79da2b1e4/src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs#L196

rynowak:
Yeah, that's right, it's supposed to be used at L198 to avoid an allocation

PR #109, opened 2018-05-04
Adopt 6.0.0 builds of polly

rynowak:
The polly packages have dropped the -Signed suffix because now they are
only offering the signed variant.

This is the version we will need to use in RTM


rynowak:
This is a dry run while we wait for the RTM versions of these packages.

Issue #110, opened 2018-05-12
Support direct creation of HttpMessageHandler instances?


martincostello:
_Related to IdentityServer/IdentityServer4#2311_

A number of libraries provide ways to interact with their internal HTTP calls via allowing the integrator to specify an `HttpMessageHandler` implementation to use. Examples include:
  1. [`RemoteAuthenticationOptions.BackchannelHttpHandler`](https://github.com/aspnet/Security/blob/f8b4f4c620e2faaeec262d302df143068c8d0b33/src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationOptions.cs#L74) used by various `Microsoft.AspNetCore.Authentication.*` packages.
  1. [`IdentityServerAuthenticationOptions.JwtBackChannelHandler`](https://github.com/IdentityServer/IdentityServer4.AccessTokenValidation/blob/3fd88d9ae73734504e269b08901779a7bac1484f/src/IdentityServer4.AccessTokenValidation/IdentityServerAuthenticationOptions.cs#L100) which is part of [`IdentityServer4.AccessTokenValidation`](https://github.com/IdentityServer/IdentityServer4.AccessTokenValidation).

From looking at the [`DefaultHttpClientFactory`](https://github.com/aspnet/HttpClientFactory/blob/9edb6d12d039efdda6a8038227b146bfd12f7ab5/src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs#L117-L118) implementation, it seems like there's currently no way to be able to create an `HttpMessageHandler` directly while retaining the named-configuration, cycling, logging etc. capabilities/behaviours gained when using `IHttpClientFactory` to create `HttpClient` instances.

As a very rough pseudo-implementation, I think what would allow easy interop for application developers to leverage the HttpClientFactory features with existing libraries would be something like this:

```csharp
public interface IHttpMessageHandlerFactory
{
    HttpMessageHandler CreateHandler(string name);
}
```

Which could then be implemented in `DefaultHttpClientFactory` by something like:

```csharp
public HttpMessageHandler CreateHandler(string name)
{
    if (name == null)
    {
        throw new ArgumentNullException(nameof(name));
    }

    var entry = _activeHandlers.GetOrAdd(name, _entryFactory).Value;
    var handler = new LifetimeTrackingHttpMessageHandlerWrapper(entry.Handler);

    StartHandlerEntryTimer(entry);

    return handler;
}

[...]

PR #118, opened 2018-05-19
Add HttpMessageHandler factory

martincostello:
Add a simple abstraction to allow leveraging the configurability of `IHttpClientFactory` for creating `HttpMessageHandler` instances that use the same configured message handler chains as are available for `HttpClient` instances.

I've just piggybacked the `DefaultHttpClientFactory` class to provide the implementation, with an extension method and a private class to wrap the message handler so it's "dispose-proof" like the handlers used in the `HttpClient` instances are.

The primary use-case is for interop with existing .NET libraries that use `HttpMessageHandler` as their HTTP abstraction rather than `HttpClient`, so that features like logging and resilience can be lit-up for these scenarios.

Relates to #110.

### DIFF
--- a/src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
+++ b/src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
@@ -117,15 +117,9 @@ namespace Microsoft.Extensions.Http
 
         public HttpClient CreateClient(string name)
         {
-            if (name == null)
-            {
-                throw new ArgumentNullException(nameof(name));
-            }
+            HttpMessageHandler handler = CreateHandler(name);
 
-            var entry = _activeHandlers.GetOrAdd(name, _entryFactory).Value;
-            var client = new HttpClient(entry.Handler, disposeHandler: false);
-
-            StartHandlerEntryTimer(entry);
+            var client = new HttpClient(handler, disposeHandler: false);
 
             var options = _optionsMonitor.Get(name);
             for (var i = 0; i < options.HttpClientActions.Count; i++)
--- a/src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
+++ b/src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
@@ -138,11 +138,10 @@ namespace Microsoft.Extensions.Http
             }
 
             var entry = _activeHandlers.GetOrAdd(name, _entryFactory).Value;
-            var handler = new LifetimeTrackingHttpMessageHandlerWrapper(entry.Handler);
 
             StartHandlerEntryTimer(entry);
 
-            return handler;
+            return entry.Handler;
         }
 
         // Internal for tests
@@ -377,21 +376,6 @@ namespace Microsoft.Extensions.Http
             public static void HandlerExpired(ILogger logger, string clientName, TimeSpan lifetime)
             {
                 _handlerExpired(logger, lifetime.TotalMilliseconds, clientName, null);
-            }
-        }
-
-        // Wraps the inner handler in the same manner as HttpClient so that the caller can
-        // dispose of the handler freely without affecting it being shared with other clients
-        private sealed class LifetimeTrackingHttpMessageHandlerWrapper : DelegatingHandler
-        {
-            internal LifetimeTrackingHttpMessageHandlerWrapper(LifetimeTrackingHttpMessageHandler handler)
-                : base(handler)
-            {
-            }
-
-            protected override void Dispose(bool disposing)
-            {
-                // The lifetime of the LifetimeTrackingHttpMessageHandler is tracked separately
             }
         }
     }
--- a/src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
+++ b/src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
@@ -117,8 +117,12 @@ namespace Microsoft.Extensions.Http
 
         public HttpClient CreateClient(string name)
         {
-            HttpMessageHandler handler = CreateHandler(name);
+            if (name == null)
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
 
+            var handler = CreateHandler(name);
             var client = new HttpClient(handler, disposeHandler: false);
 
             var options = _optionsMonitor.Get(name);
--- a/src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
+++ b/src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
@@ -15,7 +15,7 @@ using Microsoft.Extensions.Options;
 
 namespace Microsoft.Extensions.Http
 {
-    internal class DefaultHttpClientFactory : IHttpClientFactory
+    internal class DefaultHttpClientFactory : IHttpClientFactory, IHttpMessageHandlerFactory
     {
         private readonly ILogger _logger;
         private readonly IServiceProvider _services;
@@ -134,6 +134,21 @@ namespace Microsoft.Extensions.Http
             }
 
             return client;
+        }
+
+        public HttpMessageHandler CreateHandler(string name)
+        {
+            if (name == null)
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            var entry = _activeHandlers.GetOrAdd(name, _entryFactory).Value;
+            var handler = new LifetimeTrackingHttpMessageHandlerWrapper(entry.Handler);
+
+            StartHandlerEntryTimer(entry);
+
+            return handler;
         }
 
         // Internal for tests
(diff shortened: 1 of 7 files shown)


rynowak:
Just responding here for anyone else interested. There's a discussion about this going on at https://github.com/aspnet/HttpClientFactory/issues/110

martincostello:
@rynowak What sort of timescale do you need me to get this PR branch rebased for?

rynowak:
I plan to merge some kind of solution to this during the next month. Give me a few days and I'll review this and get back to you with notes. Currently thinking on this and some of the issues folks have been having with DI lifetimes. I want to understand the solution to that problem first.

rynowak:
Thanks for the patience @martincostello - I think #155 is a solution that all are going to happy with re: lifetime issues, so I'm going to give feedback on this PR assuming that #155 is merged.

rynowak:
My hope is that we could get this change in some time this week, (any time this month will likely make 2.2.0-preview2)

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
```diff
@@ -344,5 +359,20 @@ public static void HandlerExpired(ILogger logger, string clientName, TimeSpan li
                 _handlerExpired(logger, lifetime.TotalMilliseconds, clientName, null);
             }
         }
+
+        // Wraps the inner handler in the same manner as HttpClient so that the caller can
+        // dispose of the handler freely without affecting it being shared with other clients
+        private sealed class LifetimeTrackingHttpMessageHandlerWrapper : DelegatingHandler
```
rynowak: I'd prefer if we could bake this behavior into `LifetimeTrackingHttpMessageHandler` rather than creating an extra object. It seems like we could do this there with no impact on the existing behavior. Thoughts?

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
```diff
@@ -128,6 +128,21 @@ public HttpClient CreateClient(string name)
...
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            var entry = _activeHandlers.GetOrAdd(name, _entryFactory).Value;
+            var handler = new LifetimeTrackingHttpMessageHandlerWrapper(entry.Handler);
+
+            StartHandlerEntryTimer(entry);
+
+            return handler;
```
rynowak: Should `CreateClient` call into this?

src/Microsoft.Extensions.Http/IHttpMessageHandlerFactory.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// instance. Callers may cache the returned <see cref="HttpMessageHandler"/> instance indefinitely or surround
+        /// its use in a <langword>using</langword> block to dispose it when desired.
+        /// </para>
+        /// <para>
+        /// The default <see cref="IHttpMessageHandlerFactory"/> implementation may cache the underlying
+        /// <see cref="HttpMessageHandler"/> instances to improve performance.
+        /// </para>
+        /// </remarks>
+        HttpMessageHandler CreateHandler(string name);
+    }
```
rynowak: I'm wondering if there is demand for the ability to create an 'untracked' handler. Then you can use the factory to configure a pipeline, but still manage its lifetime explicitly. 

I'm going to review of the issues and see how much demand this for this.

src/Microsoft.Extensions.Http/IHttpMessageHandlerFactory.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <summary>
+        /// Creates and configures an <see cref="HttpMessageHandler"/> instance using the configuration that corresponds
+        /// to the logical name specified by <paramref name="name"/>.
+        /// </summary>
+        /// <param name="name">The logical name of the message handler to create.</param>
+        /// <returns>A new <see cref="HttpMessageHandler"/> instance.</returns>
+        /// <remarks>
+        /// <para>
+        /// Each call to <see cref="CreateHandler(string)"/> is guaranteed to return a new <see cref="HttpMessageHandler"/>
+        /// instance. Callers may cache the returned <see cref="HttpMessageHandler"/> instance indefinitely or surround
```
rynowak: Do you think returning a new instance each time is an important detail? We do this in the client factory because the `HttpClient` is mutable, but I don't think there's much value to it here.

src/Microsoft.Extensions.Http/IHttpMessageHandlerFactory.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// its use in a <langword>using</langword> block to dispose it when desired.
```
rynowak: It might make sense to explain that the factory manages the lifetime of the handler, disposing it might have no effect.

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
martincostello: Sure, I can look into doing that.

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
martincostello: Good point. I think I just did a minimal "don't touch existing stuff" approach as a first pass.

src/Microsoft.Extensions.Http/IHttpMessageHandlerFactory.cs
martincostello: Maybe this method could be something like this?

```csharp
HttpMessageHandler CreateHandler(string name, bool trackLifetime);
```

Then the caller has the option of either, and the existing plumbing for `HttpClient` can call it with `trackLifetime` as `true`?

src/Microsoft.Extensions.Http/IHttpMessageHandlerFactory.cs
martincostello: Fair point. I think this just got copied in the process of my first attempt at adding this.

src/Microsoft.Extensions.Http/IHttpMessageHandlerFactory.cs
martincostello: Fair.

martincostello:
Thanks for the feedback @rynowak - I'll try and do an initial pass at fixing this up this evening.

src/Microsoft.Extensions.Http/IHttpMessageHandlerFactory.cs
rynowak: Let me think a little bit more about the design of this. I'm ok with you proceeding with the API design as it is right now, but I might change it later.

martincostello:
@rynowak I've rebased onto `release/2.2` and made a first pass at addressing your feedback.

src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs
```diff
@@ -117,15 +117,9 @@ internal class DefaultHttpClientFactory : IHttpClientFactory
...
         public HttpClient CreateClient(string name)
         {
-            if (name == null)
-            {
-                throw new ArgumentNullException(nameof(name));
-            }
-
-            var entry = _activeHandlers.GetOrAdd(name, _entryFactory).Value;
-            var client = new HttpClient(entry.Handler, disposeHandler: false);
+            HttpMessageHandler handler = CreateHandler(name);
```
rynowak: We should still null-check here, generally we validate arguments as close as possible to user code (even if it's redundant).

Also needs some var.

test/Microsoft.Extensions.Http.Test/DependencyInjection/HttpClientFactoryServiceCollectionExtensionsTest.cs
```diff
@@ -518,6 +538,57 @@ public void AddHttpClient_WithTypedClientAndImplementation_AndServiceDelegate_Co
...
+
+            var services = serviceCollection.BuildServiceProvider();
+
+            var clientFactory = services.GetRequiredService<IHttpClientFactory>();
+            var handlerFactory = services.GetRequiredService<IHttpMessageHandlerFactory>();
+
+            // Assert
+            Assert.Same(clientFactory, handlerFactory);
+        }
+
```
rynowak: I'd also like to see a unit test somewhere that verifies that disposing a handler returned from the factory doesn't dispose the handler chain. I don't doubt that the code is correct, but I want a test since that's a very important detail and we need to preserve it.

martincostello: I've added another test case to `DefaultHttpClientFactoryTest` where there was already one to test the disposable behaviour of the `HttpClient`.

martincostello:
Thanks @rynowak! 😃

PR #122, opened 2018-05-28
Fix typo

jo-ninja:
Fix typo


jo-ninja:
Found another one. 

https://github.com/aspnet/HttpClientFactory/blob/9a393fa4ce189fee3eaefe12ba2ad8a3dc06c752/src/Microsoft.Extensions.Http/HttpMessageHandlerBuilder.cs#L25

Is it fine to change `instructure` => `infrastructure`?

rynowak:
Yes, that's right

rynowak:
I'll take care of fixing that other one. Thanks!

PR #152, opened 2018-08-07
Fix #151 - allow multiple generic typed clients

### LINKED ISSUE #151: Multiple generic typed clients will only get last added HttpClient
Let's say I have a generic class that needs a uniquely configured HttpClient injected into it.

```
public class ApiInvoker<TService>
{
    public HttpClient Client { get; }
    public ApiInvoker(HttpClient client) 
    {
        Client = client; //should be fully configured already
    }
    // Code that uses TService and Client
}
```
In ConfigureServices, I register several ApiInvokers for different services.

```
services.AddHttpClient<ApiInvoker<IServiceA>>(c => { c.BaseAddress = new Uri("http://aa.aa"); });
services.AddHttpClient<ApiInvoker<IServiceB>>(c => { c.BaseAddress = new Uri("http://bb.bb"); });
services.AddHttpClient<ApiInvoker<IServiceC>>(c => { c.BaseAddress = new Uri("http://cc.cc"); });
```

The expected behavior is that, when injected:
ApiInvoker<IServiceA> will have an HttpClient with a BaseAddress of http://aa.aa.
ApiInvoker<IServiceB> will have an HttpClient with a BaseAddress of http://bb.bb.
ApiInvoker<IServiceC> will have an HttpClient with a BaseAddress of http://cc.cc.

Instead, the outcome is:
ApiInvoker<IServiceA> has an HttpClient with a BaseAddress of http://cc.cc.
ApiInvoker<IServiceB> has an HttpClient with a BaseAddress of http://cc.cc.
ApiInvoker<IServiceC> has an HttpClient with a BaseAddress of http://cc.cc.

I believe this issue is caused by this line:
https://github.com/aspnet/HttpClientFactory/blob/8fa9de4a314bd1ac7ae6a7302a4f5ff508a43763/src/Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs#L200

In this example, the name of these separate Typed Clients are all **ApiInvoker`1**. This causes each subsequent registration to replace the one above it.

There is an overload that will allows a workaround. You can explicitly set a name for typed clients, so that you can also use them as named clients. If you set each of the names uniquely, the typed clients will not replace each other.
```
services.AddHttpClient<ApiInvoker<IServiceA>>("unique name A", c => { ... });
services.AddHttpClient<ApiInvoker<IServiceB>>("unique name B", c => { ... });
services.AddHttpClient<ApiInvoker<IServiceC>>("unique name C", c => { ... });
```

While this works in the short term, this puts the burden on me to set unnecessary names on clients I'll never need to call by name. If name is strictly required under the hood, it should just generate a GUID for the name instead of using the type name. 

Aside from the naming collision it causes, it could potentially be problematic that an HttpClientFactory could then create named clients that were never specified as named clients. I can call CreateClient("ApiInvoker`1") and get a client I never explicitly registered to that name.

Thanks!

rynowak:
Addresses a usability issue where typed clients with generic type
parameters didn't work well. This is broken because Type.Name always
uses the metadata name 'MyClientType`1' instead of the name that the
user typed in code.

This change uses some existing infrastructure we have to 'pretty print'
the type name.

The subtle breaking change here is that if you were using
`services.Configure(typeof(MyClient<>).Name, ...)` this will no longer
work. Doing these operations through the builder we provide will work
just fine, so I'm not terribly concerned.

### DIFF
--- a/src/Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs
+++ b/src/Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs
@@ -5,6 +5,7 @@ using System;
 using System.Net.Http;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Http;
+using Microsoft.Extensions.Internal;
 
 namespace Microsoft.Extensions.DependencyInjection
 {
@@ -183,9 +184,6 @@ namespace Microsoft.Extensions.DependencyInjection
         /// can be retrieved from <see cref="IServiceProvider.GetService(Type)" /> (and related methods) by providing
         /// <typeparamref name="TClient"/> as the service type. 
         /// </para>
-        /// <para>
-        /// Use <see cref="Options.Options.DefaultName"/> as the name to configure the default client.
-        /// </para>
         /// </remarks>
         public static IHttpClientBuilder AddHttpClient<TClient>(this IServiceCollection services)
             where TClient : class
@@ -197,7 +195,8 @@ namespace Microsoft.Extensions.DependencyInjection
 
             AddHttpClient(services);
 
-            var builder = new DefaultHttpClientBuilder(services, typeof(TClient).Name);
+            var name = TypeNameHelper.GetTypeDisplayName(typeof(TClient), fullName: false);
+            var builder = new DefaultHttpClientBuilder(services, name);
             builder.AddTypedClient<TClient>();
             return builder;
         }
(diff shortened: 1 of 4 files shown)


src/Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs
```diff
@@ -183,9 +184,6 @@ public static IHttpClientBuilder AddHttpClient(this IServiceCollection services,
         /// can be retrieved from <see cref="IServiceProvider.GetService(Type)" /> (and related methods) by providing
         /// <typeparamref name="TClient"/> as the service type. 
         /// </para>
-        /// <para>
-        /// Use <see cref="Options.Options.DefaultName"/> as the name to configure the default client.
-        /// </para>
```
rynowak: This wasn't correct. This paragraph is only supposed to be present on the overloads that accept the name as a string.

PR #154, opened 2018-08-12
Add retries to flaky test

rynowak:


test/Microsoft.Extensions.Http.Test/DefaultHttpClientFactoryTest.cs
```diff
@@ -402,9 +402,22 @@ public async Task Factory_CleanupCycle_DisposesLiveHandler()
...
+                GC.Collect();
+                GC.WaitForPendingFinalizers();
+                GC.Collect();
+
+                if (cleanupEntry.CanDispose)
+                {
+                    break;
+                }
+
+                Thread.Sleep(TimeSpan.FromSeconds(1));
```
pranavkm: `await Task.Delay(TimeSpan.FromSeconds(1))`

rynowak:
I might have actually figured this out BTW. 

I think there was a codegen difference for linux on debug, that was causing the `HttpClient` instance to be held in a field of the state machine. So I added a stack frame 👍

Issue #134, opened 2018-06-14
Creating a scoped service with an HttpClientFactory


twsouthwick:
I really like the idea of HttpClientFactory, but have found issues that the convenience methods assume you have singleton instances of the services required. For instance, the following fails:

```csharp
public void ConfigureServices(IServiceCollection services)
{
    services.AddScoped<Test2>(); // Works if I switch this to AddSingleton
    services.AddHttpClient<Test1>();
    services.AddMvc().SetCompatibilityVersion(CompatibilityVersion.Version_2_1);
}

public class Test1
{
   public Test1(HttpClient client, Test2 test2)
   {
   }
}

public class Test2 { }
```

This results in the following:

```
InvalidOperationException: Cannot resolve scoped service 'WebApplication1.Startup+Test2' from root provider.
```

Can there be a way to set the lifetime of the scope? In this example, I want to set `Test1` as a singleton.

twsouthwick:
I found that I can do the following at it starts working:

```csharp
var type = typeof(ITypedHttpClientFactory<>).Assembly.DefinedTypes.Where(t => t.Name.Contains("DefaultTypedHttpClientFactory")).Single();
services.AddScoped(typeof(ITypedHttpClientFactory<>), type);
```

Is this ok to do?

twsouthwick:
So this?

```csharp
var type = typeof(ITypedHttpClientFactory<>).Assembly.DefinedTypes.Single(t => t.Name.Contains("DefaultTypedHttpClientFactory"));
services.AddTransient(typeof(ITypedHttpClientFactory<>), type);
```

martincostello:
What I meant was to register Test2 as a transient, rather than scoped, dependency. Alternatively it _may_ be related to being a nested type. I’m not at a computer at the moment so I don’t have an IDE to hand to go digging.

theezak:
Injecting handlers is however subject to a memory leak (see other issue opened on that topic). The dependency container will keep a reference to the handler because the handler implements IDisposable. That prevents the handler from being released for garbage collection. A way around it is using a container that allows you to disable IDisposable tracking (for example ExternallyOwned () on autofac)

brockallen:
I have the same issue. I am injecting a DelegatingHandler which itself has an indirect dependency on a DbContext from DI which is scoped. This blocks the ability to use AddHttpMessageHandler for that scenario.

rynowak:
Assigning myself to take a look at this.

[...]

PR #155, opened 2018-08-13
Fix #124 and #134

### LINKED ISSUE #124: IDisposable and transient handlers that are injected
Hi,

There seems to be a conflict between the cached lifetime of a handler and the container calling IDisposable. Even when transient the container will dispose of the handler before the handler expires in the cache. We are running Autofac, but there are other containers with similar behavior.

rynowak:
This is a rework of the http client factory interacts with dependency
injection when creating a message handler. Since the http client factory
manages the lifetimes of message handlers, it's not appropriate for a
message handler and related services to share lifetimes with the
application or 'current request' scope.

Now the factory creates a scope each time it creates a handler, and then
disposes the scope when the handler is disposed. This allows the handler
building process to resolve scoped services and also to prevent
allocating lots and lots of disposable transients in the application
global scope.

The fix for #134 also required a rework of some details of
`DefaultTypedClientFactory<>`.

### DIFF
--- a/src/Microsoft.Extensions.Http/DefaultTypedHttpClientFactory.cs
+++ b/src/Microsoft.Extensions.Http/DefaultTypedHttpClientFactory.cs
@@ -11,11 +11,11 @@ namespace Microsoft.Extensions.Http
     internal class DefaultTypedHttpClientFactory<TClient> : ITypedHttpClientFactory<TClient>
     {
         private readonly static Func<ObjectFactory> _createActivator = () => ActivatorUtilities.CreateFactory(typeof(TClient), new Type[] { typeof(HttpClient), });
-        private readonly IServiceProvider _services;
+        private static ObjectFactory _activator;
+        private static bool _initialized;
+        private static object _lock;
 
-        private ObjectFactory _activator;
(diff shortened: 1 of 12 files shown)


src/Microsoft.Extensions.Http/DefaultTypedHttpClientFactory.cs
```diff
@@ -11,11 +11,11 @@ namespace Microsoft.Extensions.Http
     internal class DefaultTypedHttpClientFactory<TClient> : ITypedHttpClientFactory<TClient>
     {
         private readonly static Func<ObjectFactory> _createActivator = () => ActivatorUtilities.CreateFactory(typeof(TClient), new Type[] { typeof(HttpClient), });
-        private readonly IServiceProvider _services;
+        private static ObjectFactory _activator;
+        private static bool _initialized;
+        private static object _lock;
```
rynowak: This fix is likely not good enough. One of the issues with statics + ActivatorUtilities is that the delegate we create is only value for that particular service provider configuration.

We'll need to move to a model where we have a singleton cache + these objects as transients.

Issue #171, opened 2018-09-02
Don't capture the ExecutionContext in Timers created for expiry


davidfowl:
Any lazy activation of timers will capture the ExecutionContext. Combining this with a lazy initialization of the HttpClient and the handler graph may end up holding onto AsyncLocals for longer than expected. This *could* end up looking like a memory leak. 

Here's an example of the `HttpContext` being captured by the timer inside of the ActiveHandlerTrackingEntry's timer's ExecutionContext.

```
->  000001c2ef4e50e0 Microsoft.Extensions.Http.DefaultHttpClientFactory
->  000001c2ef4e5608 System.Collections.Concurrent.ConcurrentDictionary`2[[System.String, System.Private.CoreLib],[System.Lazy`1[[Microsoft.Extensions.Http.ActiveHandlerTrackingEntry, Microsoft.Extensions.Http]], System.Private.CoreLib]]
->  000001c2ef4e57b8 System.Collections.Concurrent.ConcurrentDictionary`2+Tables[[System.String, System.Private.CoreLib],[System.Lazy`1[[Microsoft.Extensions.Http.ActiveHandlerTrackingEntry, Microsoft.Extensions.Http]], System.Private.CoreLib]]
->  000001c2ef4e56a8 System.Collections.Concurrent.ConcurrentDictionary`2+Node[[System.String, System.Private.CoreLib],[System.Lazy`1[[Microsoft.Extensions.Http.ActiveHandlerTrackingEntry, Microsoft.Extensions.Http]], System.Private.CoreLib]][]
->  000001c2ef5345c0 System.Collections.Concurrent.ConcurrentDictionary`2+Node[[System.String, System.Private.CoreLib],[System.Lazy`1[[Microsoft.Extensions.Http.ActiveHandlerTrackingEntry, Microsoft.Extensions.Http]], System.Private.CoreLib]]
->  000001c2ef534578 System.Lazy`1[[Microsoft.Extensions.Http.ActiveHandlerTrackingEntry, Microsoft.Extensions.Http]]
->  000001c2ef5365f8 Microsoft.Extensions.Http.ActiveHandlerTrackingEntry
->  000001c2ef5367a0 System.Threading.Timer
->  000001c2ef536838 System.Threading.TimerHolder
->  000001c2ef5367b8 System.Threading.TimerQueueTimer
->  000001c2ef533470 System.Threading.ExecutionContext
[...]

PR #172, opened 2018-09-02
Don't capture asynclocals on timers

benaadams:
Workaround for https://github.com/dotnet/corefx/issues/26523

Resolves https://github.com/aspnet/HttpClientFactory/issues/171

/cc @davidfowl


davidfowl:
Make a helper that does this and call it from both places?

rynowak:
@benaadams thanks! I'm going to merge this and then take a stab at de-duping the codez

PR #191, opened 2018-10-14
Extract IHttpClientFactory, IHttpMessageHandlerFactory and ITypedHttpClientFactory<TClient> to an abstractions assembly.

paulomorgado:
These abstractions are also useful for non-server applications.

Splitting them allows for libraries that do not depend on server ASP.NET Core.

Allows also for other implementations without taking a dependency on the provided implementation.


rynowak:
Hi @paulomorgado - have you considered using [](https://github.com/aspnet/Hosting/blob/master/samples/GenericHostSample/ProgramHelloWorld.cs)generic host for this? The client factory intentionally don't depend on anything from ASP.NET Core - just the Microsoft.Extensions libraries.

Do you plan to create and ship an alternative implementation of this for your use case?

paulomorgado:
@rynowak, my components/libraries can run on console applications, windows services, system.web and any other runtime.

But, besides that, I like the model of having abstractions assemblies used by other Microsoft.Extensions Libraries. I use them a lot and I haven't written an ASP.NET Core or .NET Core application ever.

rynowak:
> have you considered using generic host for this? The client factory intentionally don't depend on anything from ASP.NET Core - just the Microsoft.Extensions libraries.

> Do you plan to create and ship an alternative implementation of this for your use case?

paulomorgado:
> have you considered using generic host for this? The client factory intentionally don't depend on anything from ASP.NET Core - just the Microsoft.Extensions libraries.

Not really compatible with System.Web. Is it?

> Do you plan to create and ship an alternative implementation of this for your use case?

I already have implemented it. It just take a bit of remaining to become `IHttpClientFactory` and `IHttpMessageHandlerFactory`.

rynowak:
> Not really compatible with System.Web. Is it?

Yes, absolutely. The generic host libraries target netstandard2.0 and net461 - they can be used to bootstrap the Microsoft.Extensions libraries including logging and configuration into any .NET process.

paulomorgado:
Cool! Do you have an example on how I can set up a generic host on top of System.Web (ASP.NET WebForms, ASP.NET MVC and ASP.NET Web API) hosting?

rynowak:
Here's a general purpose example: https://github.com/aspnet/Hosting/blob/master/samples/GenericHostSample/ProgramFullControl.cs - I think you'd primarily be interested in the services configuration part for HttpClient Factory.

If you're on System.Web then you'd want to call `StartAsync` and `StopAsync` from the relevant events in Global.asax.cs (I don't remember the event names)

paulomorgado:
I'll have a look at it.

It does look like a very convoluted way of using this for System.Web (or OWIN, for that matter).

natemcmaster:
Hey @paulomorgado, I'm going to archive this repo later today which makes it read-only. HttpClientFactory is still under active development in this repo now: https://github.com/aspnet/Extensions/. If you'd like to continue this work, you'll have to rebase your changes on that repo and create a new PR.

Thanks,
Nate

PR #197, opened 2018-10-19
Fix #194 Null Reference in ActiveHandlerTrackingEntry

### LINKED ISSUE #194: NullReferenceException in ActiveHandlerTrackingEntry with 2.2.0 preview 3
I don't yet have a repro, but I've just updated an ASP.NET Core web app that uses IHttpClientFactory to 2.2.0 preview 3, and I've got some in-memory self-hosted tests failing because of the following exception causing `dotnet test` to terminate:

```
[18:18:48][dotnet test] The active test run was aborted. Reason: Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
[18:18:48][dotnet test]    at Microsoft.Extensions.Http.ActiveHandlerTrackingEntry.Timer_Tick()
[18:18:48][dotnet test]    at Microsoft.Extensions.Http.ActiveHandlerTrackingEntry.<>c.<.cctor>b__22_0(Object s)
[18:18:48][dotnet test]    at System.Threading.TimerQueueTimer.CallCallback()
[18:18:48][dotnet test]    at System.Threading.TimerQueueTimer.Fire()
[18:18:48][dotnet test]    at System.Threading.TimerQueue.FireNextTimers()
[18:18:48][dotnet test]    at System.Threading.TimerQueue.AppDomainTimerCallback(Int32 id)
```

I guess that one of either `_timer` or `_callback` is somehow `null`:

https://github.com/aspnet/HttpClientFactory/blob/5255a24fe5bc5585e6fda2abaf78ae75936fa30b/src/Microsoft.Extensions.Http/ActiveHandlerTrackingEntry.cs#L75-L87

rynowak:
This issue here is that we were allowing multiple timers to start
because we never set `_timerInitialized`. This is a regression during
the milestone from introducing NonCapturingTimer.

### DIFF
--- a/src/Microsoft.Extensions.Http/ActiveHandlerTrackingEntry.cs
+++ b/src/Microsoft.Extensions.Http/ActiveHandlerTrackingEntry.cs
@@ -69,6 +69,7 @@ namespace Microsoft.Extensions.Http
 
                 _callback = callback;
                 _timer = NonCapturingTimer.Create(_timerCallback, this, Lifetime, Timeout.InfiniteTimeSpan);
+                _timerInitialized = true;
             }
         }
 
@@ -79,10 +80,13 @@ namespace Microsoft.Extensions.Http
 
             lock (_lock)
             {
-                _timer.Dispose();
-                _timer = null;
+                if (_timer != null)
+                {
+                    _timer.Dispose();
+                    _timer = null;
 
-                _callback(this);
+                    _callback(this);
+                }
             }
         }
     }


src/Microsoft.Extensions.Http/ActiveHandlerTrackingEntry.cs
```diff
@@ -79,10 +80,13 @@ private void Timer_Tick()
 
             lock (_lock)
             {
-                _timer.Dispose();
-                _timer = null;
+                if (_timer != null)
```
rynowak, line 83: Just some extra non-crashing here. We already assert if we get here and this is null.

pranavkm: Btw, does this type need a finalizer? We did it with file providers to work around unit \ functional tests that weren't disposing these types

pranavkm (approved):

src/Microsoft.Extensions.Http/ActiveHandlerTrackingEntry.cs
rynowak: This isn't ever exposed to users, it's an implementation detail of the factory.

